PR #6, opened 2018-10-19
Ensure only one open connection to API node

jontey:
When rest gateway receives a burst of transactions on startup, it will open a new connection for each request as `connectionService` has not saved the connection while waiting for the verification to complete.


Vektrat:
Hello, we are holding this PR because forcing connections on startup seems a bit like a hack, we are trying to come up with a cleaner solution within the connectionService

Vektrat:
Closing to clean up issues list, will consider again when cleaning up code, but definitely minor and no obvious clean strategy to approach this atm ^^ thanks for the heads up though

PR #20, opened 2019-04-10
Added a server info endpoint providing version number details

Vektrat:
I think this will help with reporting errors and the number of confusions that have sprouted lately with the increased dev activity

@dgarcia360 probably of interest for you for documentation, @evias also a "tool" you might find useful

### DIFF
--- a/catapult-sdk/src/model/ModelSchemaBuilder.js
+++ b/catapult-sdk/src/model/ModelSchemaBuilder.js
@@ -153,6 +153,10 @@ class ModelSchemaBuilder {
 				communicationTimestamps: { type: ModelType.object, schemaName: 'communicationTimestamps' }
 			},
 
+			serverInfo: {
+
+			},
+
 			storageInfo: {
 			}
 
--- a/catapult-sdk/src/model/ModelSchemaBuilder.js
+++ b/catapult-sdk/src/model/ModelSchemaBuilder.js
@@ -154,7 +154,6 @@ class ModelSchemaBuilder {
 			},
 
 			serverInfo: {
-
 			},
 
 			storageInfo: {
--- a/rest/src/routes/diagnosticRoutes.js
+++ b/rest/src/routes/diagnosticRoutes.js
@@ -20,15 +20,11 @@
 
 const routeResultTypes = require('./routeResultTypes');
 const routeUtils = require('./routeUtils');
+const { version: restVersion } = require('../../package.json');
+const { version: sdkVersion } = require('../../../catapult-sdk/package.json');
 
 module.exports = {
 	register: (server, db) => {
-		server.get('/diagnostic/storage', (req, res, next) =>
-			db.storageInfo().then(storageInfo => {
-				res.send({ payload: storageInfo, type: routeResultTypes.storageInfo });
-				next();
-			}));
-
 		server.get('/diagnostic/blocks/:height/limit/:limit', (req, res, next) => {
 			const parseUint = paramName => routeUtils.parseArgument(req.params, paramName, 'uint');
 			const height = parseUint('height');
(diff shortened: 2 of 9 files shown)


dgarcia360:
Developers could find confusing to have both ``node/info`` and ``diagnostic/serverInfo`` endpoints.

Consider one of the following improvements:

1. Move the new endpoint to ``/node/version``. To have a better experience, we could access the versions using an identifier instead of looking an ordered array.

    {
            "node": [nodeVersion](https://github.com/nemtech/catapult-server/blob/master/resources/config-node.properties#L51),
            "rest": restVersion,
            "sdk":sdkVersion
    }

2. Reuse``node/info`` endpoint. This already has a version field, which refers to the version defined in the node configuration parameters. The downside is that this change would break compatibility.

catapult-sdk/src/model/ModelSchemaBuilder.js
```diff
@@ -153,6 +153,10 @@ class ModelSchemaBuilder {
 				communicationTimestamps: { type: ModelType.object, schemaName: 'communicationTimestamps' }
 			},
 
+			serverInfo: {
+
```
Jaguar0625: match following `storageInfo ` format and don't have blank line

rest/src/routes/diagnosticRoutes.js
```diff
@@ -20,15 +20,11 @@
 
 const routeResultTypes = require('./routeResultTypes');
 const routeUtils = require('./routeUtils');
+const { version: restVersion } = require('../../package.json');
+const { version: sdkVersion } = require('../../../catapult-sdk/package.json');
```
Jaguar0625: clever 👍

rest/src/routes/diagnosticRoutes.js
```diff
@@ -39,5 +35,24 @@ module.exports = {
 				next();
 			});
 		});
+
+		server.get('/diagnostic/serverInfo', (req, res, next) => {
```
Jaguar0625: would probably prefer dropping info as /server and /storage seem more consistetn

rest/test/routes/diagnosticRoutes_spec.js
```diff
@@ -21,11 +21,72 @@
...
+			const next = () => {};
+			const res = {
+				send: response => {
+					sentResponse = response;
+				},
+				redirect: () => {
+					next();
+				}
+			};
+			// Act:
```
Jaguar0625: add newline

rest/test/routes/diagnosticRoutes_spec.js
```diff
@@ -21,11 +21,72 @@
...
+			let sentResponse;
+			const next = () => {};
+			const res = {
+				send: response => {
+					sentResponse = response;
+				},
+				redirect: () => {
+					next();
+				}
+			};
```
Jaguar0625: should really refactor these eventually ^^

rest/test/routes/diagnosticRoutes_spec.js
```diff
@@ -21,11 +21,72 @@
...
+			expect(sentResponse).to.deep.equal({
+				payload: {
+					serverInfo: {
+						restVersion,
+						sdkVersion
+					}
+				},
+				type: 'serverInfo'
+			});
+			done();
```
Jaguar0625: test is sync so i don't think you need to call done explicitly

rest/test/routes/diagnosticRoutes_spec.js
```diff
@@ -21,11 +21,72 @@
...
+			};
+			// Act:
+			const route = routes[endpointUnderTest];
+			route({}, res, next);
+
+			// Assert:
+			expect(sentResponse).to.deep.equal({
+				payload: {
+					serverInfo: {
+						restVersion,
```
Jaguar0625: doesn't matter but use same ordering in src + test

rest/test/routes/diagnosticRoutes_spec.js
```diff
@@ -21,11 +21,72 @@
...
+			const routes = {};
+			const server = {
+				get: (path, handler) => {
+					routes[path] = handler;
+				}
+			};
+
+			diagnosticRoutes.register(server, {});
+
+			let sentResponse;
```
Jaguar0625: sent -> captured ?

rest/test/routes/diagnosticRoutes_spec.js
Vektrat: yes, this is on the list of tech debt i owe you 😅

PR #24, opened 2019-04-24
Added inflation receipt

martinayora:


rest/test/plugins/db/utils/receiptsDbTestUtils.js
```diff
@@ -51,6 +51,12 @@ const createTransactionStatement = height => ({
 			version: 1,
 			type: 8517,
 			artifactId: Long.fromNumber(1234)
+		},
+		{
```
Vektrat: i think it'd be good to put a comment with each receipt type here ( // inflation, etc) for readability

Issue #27, opened 2019-05-02
Receipt - AddressResolutionStatement resolved value has wrong format.


rg911:
`resolved` value should be encoded in `Hex` format.

See. [http://c2.nem.ninja:3000/block/19/receipts](http://c2.nem.ninja:3000/block/19/receipts)

```
{"height":[19,0],
"unresolved":"91E4EDFC6EE000CEFF00000000000000000000000000000000",
"resolutionEntries":[
      {"source":{"primaryId":3,"secondaryId":0},
       "resolved":"kHiyJYtPHGBhINoJIjut3USNYP0cq7DJnw=="
      }
]
}
```

Vektrat:
@gimer @BloodyRookie was `entries` renamed to `resolutionEntries`?

BloodyRookie:
i don't remember renaming it, maybe it was like this from the beginning?

Vektrat:
nope, https://phab.nem.ninja/P210 but nevermind, updated ^^

PR #28, opened 2019-05-03
it should be resolutionEntries 

gimer:
fixes issue #27


Vektrat:
commited this to master a few minutes ago since it was a simple rename, sorry ^^

PR #32, opened 2019-05-22
Updated mosaic schema definition

### LINKED ISSUE #31: /mosaics endpoint not working (error 500)
> error 500 InternalError: value.getLowBitsUnsigned is not a function

the `/mosaics` endpoint is currently failing due to a change in how properties are represented at the database level, mosaic properties should be an object including id and value

martinayora:
Fixes #31. Now the value type passed to longToUint64 is also checked.

### DIFF
--- a/catapult-sdk/src/plugins/mosaic.js
+++ b/catapult-sdk/src/plugins/mosaic.js
@@ -54,7 +54,7 @@ const mosaicPlugin = {
 
 			height: ModelType.uint64,
 			owner: ModelType.binary,
-			properties: { type: ModelType.array, schemaName: ModelType.uint64 }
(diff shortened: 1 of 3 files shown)


catapult-sdk/src/plugins/mosaic.js
```diff
@@ -54,7 +54,7 @@ const mosaicPlugin = {
 
 			height: ModelType.uint64,
 			owner: ModelType.binary,
-			properties: { type: ModelType.array, schemaName: ModelType.uint64 }
+			properties: { type: ModelType.array, schemaName: 'mosaicDefinition.mosaicProperty' }
```
Vektrat: we will need to think something in the near future to add tests to those cases, i just saw the project does not test this in any other situation

PR #56, opened 2019-07-11
Upgraded nyc dependency to latest

### LINKED ISSUE #55: Upgrade REST dependencies
Upgrade REST dependencies to avoid warnings.

martinayora:
Fixes #55

### DIFF
--- a/rest/package.json
+++ b/rest/package.json
@@ -40,7 +40,7 @@
     "catapult-sdk": "link:../catapult-sdk",
     "coveralls": "^3.0.1",
     "mongodb": "^3.0.3",
-    "nyc": "^11.8.0",
+    "nyc": "^14.1.1",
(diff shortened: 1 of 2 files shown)


Vektrat:
the requirement was to update `mem` to >4.0 and `handlebars` to >4.1, for this, `nyc` was to be updated (since they are dependencies of this package) - when doing so, the nested `get-caller-file` also requires to be updated, problem is, this package has dropped compatibility with node9 (due to EOL status)

https://github.com/stefanpenner/get-caller-file/commit/96c6a94c8cbe251135596784bdf8a86e40c713ee

running yarn with the `--ignore-engines` check flag will bypass this check, resulting in potential runtime problems for unsupported versions (i believe it should be ok in this case, but i'd rather not use this for other possible cases)

PR #73, opened 2019-08-09
Changed transaction fields to match catbuffer

### LINKED ISSUE #69: ModificationType instead of RestrictionType
- AccountRestrictionTransaction 

`modifications[].modificationType` has been renamed to `restrictionType` for some reason. 

The key name should stay with `modificationType`

martinayora:
Fixes #69 

Now transaction fields match between catapult-rest and catbuffers.


Vektrat:
added @rg911 to the review since he's been playing a lot lately with those endpoints and may recall some other changes, just take a look at the changed names, no need to go in depth in the code steven

Issue #84, opened 2019-08-26
Add Uint64 formatter to string or hexadecimal


evias:
Following this PR nemtech/nem2-sdk-typescript-javascript#239, it is being discussed to add a formatter in REST to encode UInt64 as `string` (number or hexadecimal TBC) instead of in the form of an array of numbers.

As each SDK would need a copy of the PR mentioned above, it would make more sense to add the formatter in catapult-rest.

@Vektrat, assigned steven temporarily, can be changed obviously :+1:

Vektrat:
i never liked the array format, so sounds good to me, what opinion does the core team have on this? @Jaguar0625 @gimer @BloodyRookie

fboucquez:
Perfect!

IMHO, I don't see the performance benefit of having high/low number style arrays. SDKs need to convert them back to u64 numbers (like BigInteger in Java), it complicates the rest APIs making them less user-friendly and it complicates SDK mappings. I can understand that high/low can be super performant in catbuffer style serialization but not so sure in standard rest.

I also preferred the string/decimal numbers over hexadecimal numbers for the same reasons above. Unless there is a real performance reason to do it differently, a person should be able to understand the "difficulty" by just looking at http://167.71.13.249:3000/block/1.

Cheers
Fernando

Vektrat:
waiting for final input from @dgarcia360 , i believe to have seen a comment where he suggested different formats

rg911:
> waiting for final input from @dgarcia360 , i believe to have seen a comment where he suggested different formats

you mean this?
by @dgarcia360 - " Thinking more about it, for identifiers (mosaics and namespaces) could be hexa, but for height, amounts, duration, difficulty, scores, and timestamps could be a string."

dgarcia360:
My proposal was to treat all Uint64 arrays as strings except identifiers (mosaic and namespace ids). 

Differently to values, identifiers do not have a meaning on their own. For that reason, we can show them nicely in a shorter form encoding them as hexa. 

Also, it will be compatible with the current endpoint parameters (e.g. /mosaic/<mosaicId (hexa)>).

Vektrat:
Okay, with `Uint64` as `string` do you mean to represent a number in a string format? such as '7862759275' --> `"7862759275"`?

dgarcia360:
Yes, for example:

> UInt64 [3127188303,2974383967] --> string '12774881867138931535'  --> hexa 'B1497F5FBA651B4F'

[...]

PR #87, opened 2019-08-30
Added the uint64String formatter

Vektrat:
Required for #84 

This enhancement does not break backward compatibility, hence the early partial PR.

The idea would be to have an extra formatter option (uint64String), and update the current (uint64) one so that it formats to HEX instead of an old uint64 two-valued array. Every uint64 value output by REST would be represented in HEX, excluding a few select ones represented by uint64String (mosaic and namespace identifiers).

Open for renaming suggestions (uint64NumericString, uint64Identifier, etc)

### DIFF
--- a/catapult-sdk/src/model/ModelType.js
+++ b/catapult-sdk/src/model/ModelType.js
@@ -43,10 +43,13 @@ const ModelType = {
 	uint16: SchemaType.max + 5,
 
 	/** Schema property type indicating a uint64 value. */
-	uint64: SchemaType.max + 6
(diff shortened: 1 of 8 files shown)


Vektrat:
>     * If `uint64String` to be a generic model type, should the name `uint64` be changes accordingly with more meaningful names?
> 
>     * otherwise, why not keep the `uint64` as type and add `toString/toHex` as 'extensions'?

the current `uint64String` will be the new/default `uint64`, having the old one still available is not necessary dead code is dead ^^

PR #88, opened 2019-09-02
Updated chainStatistic schema

Vektrat:
with changes to chainStatistic schema chain endpoints, and chain related endpoints (dependent on `height`) are failing

as  discussed with @dgarcia360  we left the `/chain/info` endpoint with the full schema info (including `current` structure, and specific endpoints (`/chain/height` and `/chain/score`) with just inner `current` information

### DIFF
--- a/catapult-sdk/src/model/ModelFormatterBuilder.js
+++ b/catapult-sdk/src/model/ModelFormatterBuilder.js
@@ -39,6 +39,7 @@ class ModelFormatterBuilder {
 			'transactionWithMetadata',
 
 			'chainStatistic',
+			'chainStatisticCurrent',
 			'merkleProofInfo',
 			'nodeInfo',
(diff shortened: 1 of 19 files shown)


Vektrat:
>     * What's the rationale behind `chainStatisticCurrent`? (sorry wasn't involved in the conversation.)
> 
>     * Feel the name: `chainStatisticCurrent` a bit confusing. `currentChainStatistic` better ??

I believe it is to allow having `meta` introduced into the collection, you can see the details in the wg-standard document https://docs.google.com/document/d/1GpUR2Ct1DJ-VGVFGoVVGPoy_ndRmN2RcNJpNxZqapvo/edit#heading=h.612htqwbz33z

i prefer `chainStatisticCurrent` because it "groups" better with the other `chainStatistic` schema name

PR #95, opened 2019-09-04
Modified uint64 default formatter to string and uint64 ids to hex.

### LINKED ISSUE #84: Add Uint64 formatter to string or hexadecimal

martinayora:
Fixes #84.

### DIFF
--- a/catapult-sdk/test/model/ModelSchemaBuilder_spec.js
+++ b/catapult-sdk/test/model/ModelSchemaBuilder_spec.js
@@ -237,7 +237,6 @@ describe('model schema builder', () => {
 				'activityBucket.startHeight',
 				'activityBucket.totalFeesPaid',
 				'activityBucket.rawScore',
-				'mosaic.id',
 				'mosaic.amount',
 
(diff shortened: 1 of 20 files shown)


catapult-sdk/test/model/ModelSchemaBuilder_spec.js
```diff
@@ -231,7 +231,6 @@ describe('model schema builder', () => {
 				'account.publicKeyHeight',
 				'account.importance',
 				'account.importanceHeight',
-				'mosaic.id',
```
Vektrat: should add a new `it` group of tests for `uint64HexIdentifier` with this in it

Issue #100, opened 2019-09-06
Internal error on `account/.../multisig/graph


rg911:
Got internal error calling multisig/graph endpoint

```JSON
{
    "code": "Internal",
    "message": "Cannot read property 'forEach' of undefined"
}
```

PR #103, opened 2019-09-07
Fixed account address field name in multisig endpoint

### LINKED ISSUE #99: Cannot query account/{accountId}/multisig endpoint using address
This endpoint should take either `Address` / `PublicKey` as `AccountId`. Now it only accepts `PublicKey`

martinayora:
Fixes #99.

Based on https://github.com/nemtech/catapult-server/blob/master/scripts/mongo/mongoMultisigDbPrepare.js#L4


Vektrat:
most likely fixes #100 as well

PR #115, opened 2019-09-10
Fetch transactions by address

### LINKED ISSUE #8: Fetch transactions by Address
In the endpoints, it should be possible using the address as well.

- http://localhost:3000/account/{address|publicKey}/transactions
- http://localhost:3000/account/{address|publicKey}/transactions/outgoing
- http://localhost:3000/account/{address|publicKey}/transactions/incoming

### Motivation

Often, you don't know the public key, just the address. Go to the address info, get the public key and then fetch the transactions can be done, but it's not as direct as being able to fetch directly using the Address.

martinayora:
Fixes #8.

### DIFF
--- a/rest/src/routes/accountRoutes.js
+++ b/rest/src/routes/accountRoutes.js
@@ -22,9 +22,6 @@ const routeResultTypes = require('./routeResultTypes');
 const routeUtils = require('./routeUtils');
 const AccountType = require('../plugins/AccountType');
 const errors = require('../server/errors');
-const catapult = require('catapult-sdk');
-
-const { convert } = catapult.utils;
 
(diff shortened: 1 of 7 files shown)


rest/src/routes/accountRoutes.js
```diff
@@ -59,13 +56,21 @@ module.exports = {
 			{ dbPostfix: 'Unconfirmed', routePostfix: '/unconfirmed' }
 		];
 
+		const accountIdToPublicKey = (type, accountId) => {
+			if (AccountType.publicKey === type)
+				return Promise.resolve(accountId);
```
Vektrat: newline

PR #120, opened 2019-09-16
Fixed metadata schemas so that value is treated as string

### LINKED ISSUE #113: Error when announcing MosaicMetadataTransaction
When announcing MetadataTransaction I got below server error

```
"stack":"TypeError: input.forEach is not a function\n    at Object.uint8ToHex (/Users/stevenliu/Documents/src/NEM/catapult-rest/catapult-sdk/_build/utils/convert.js:103:9)\n
```
Seems like the error is recent changes related as the metadata tx worked before.

It is also worth checking the other 2 metadata txs

Vektrat:
Fixes #113

### DIFF
--- a/catapult-sdk/src/plugins/metadata.js
+++ b/catapult-sdk/src/plugins/metadata.js
@@ -39,21 +39,21 @@ const metadataPlugin = {
 		builder.addTransactionSupport(EntityType.accountMetadata, {
 			targetPublicKey: ModelType.binary,
 			scopedMetadataKey: ModelType.uint64,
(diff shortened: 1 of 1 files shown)


dgarcia360:
Is ``value`` returned as an hexa string or regular string?

Vektrat:
> Is `value` returned as an hexa string or regular string?

regular string, i believe it works like the `namespace.name`, with its size and such

but it is received as uint8array... so i'm not 100% sure it will be readable (probably yes)

Issue #78, opened 2019-08-14
add and configure throttle for api requests


Jaguar0625:
http://restify.com/docs/plugins-api/ : throttle

PR #121, opened 2019-09-16
Added options to configure throttling for the API

Vektrat:
Addresses #78 

@dgarcia360 you might be interested in tracking this in order to update config docs (when/if accepted)

### DIFF
--- a/rest/resources/rest.json
+++ b/rest/resources/rest.json
@@ -50,6 +50,11 @@
     "allowOptionalAddress": true
   },
 
+  "throttling": {
+    "burst": 20,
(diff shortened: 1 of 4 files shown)


Vektrat:
I decided to let the user opt in-out of throttling himself (by providing full config or not), if you think this is a bad idea and that it should always be activated letting the user choose the values, let me know

the provided `rest.json` values are personal suggestion, feel free to provide new ones

there's the option to throttle by username (does not apply to our project since there's no auth), and masked throttling, which i assumed is out of the question as well - leaving just IP throttling enabled

Issue #74, opened 2019-08-09
Add mosaic restriction REST endpoints


Vektrat:
Blocked by and based on: https://github.com/nemtech/nem2-openapi/issues/6

dgarcia360:
@roger  @martin An initial list of endpoints we could have to interact with [Mosaic Restrictions](http://nemtech.github.io/concepts/mosaic-restrictions.html). Could you make a PR with the final specification? Also, review the endpoints structure and list any missing additional endpoint do you think should we have.

**GlobalMosaicRestriction endpoints**

1. **GET** ``/mosaic/{mosaicId}/restrictions``

Returns an array of global mosaic restrictions assigned to the mosaic. Ideally paginated and orderable by key and type.

**AddressMosaicRestriction endpoints**

I propose to organize address mosaic restrictions under mosaics rather than accounts to avoid conflicts with account restrictions endpoints. AccountId stands for address or publicKey. If only address is possible, replace ``account/{accountId}/`` for ``../address/{address}``.

2. **GET** ``/mosaic/{mosaicId}/restrictions/account/{accountId}/``

Returns array of address mosaic restrictions for the given a ``mosaicId`` and ``targetAddress``. Ideally paginated and orderable by key.

Extra: We could have as well POST methods (/mosaic/restrictions and /mosaic/{mosaicId}/restrictions) to get different responses the same call, as we are doing with account restrictions.

PR #129, opened 2019-09-26
Add mosaic restriction endpoints REST

martinayora:
Adds mosaic restrictions endpoints, routes + db + tests.

### DIFF
--- a/rest/src/plugins/mosaicRestrictions/MosaicRestrictionsDb.js
+++ b/rest/src/plugins/mosaicRestrictions/MosaicRestrictionsDb.js
@@ -35,7 +35,7 @@ class MosaicRestrictionsDb {
 
 	/**
 	 * Retrieves mosaic restrictions of the given mosaic ids.
-	 * @param {array<object>} mosaicIds Given mosaic ids.
+	 * @param {array<module:catapult.utils/uint64~uint64>} mosaicIds Given mosaic ids.
 	 * @param {int} restrictionType Restriction type.
 	 * @returns {Promise.<array>} Mosaic restrictions.
 	 */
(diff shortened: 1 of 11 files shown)


martinayora:
All tests are successful. But because of MockServer testing utility, two paths are overlapping. 

When adding routes to server here: 
https://github.com/nemtech/catapult-rest/blob/master/rest/test/routes/utils/routeTestUtils.js#L32

the route "_/mosaic/:mosaicId/restrictions_" is beeing added both as **GET** and **POST**. And the last one is overriding the previous.

Fixing the mockserver utility will allow the tests to pass as they are now.

rest/src/plugins/mosaicRestrictions/MosaicRestrictionsDb.js
```diff
@@ -0,0 +1,74 @@
...
+	 * Creates MosaicRestrictionsDb around CatapultDb.
+	 * @param {module:db/CatapultDb} db Catapult db instance.
+	 */
+	constructor(db) {
+		this.catapultDb = db;
+	}
+
+	/**
+	 * Retrieves mosaic restrictions of the given mosaic ids.
+	 * @param {array<object>} mosaicIds Given mosaic ids.
```
Vektrat: can do this @param {Array.<module:catapult.utils/uint64~uint64>}, for the params further down as well

rest/src/plugins/mosaicRestrictions/mosaicRestrictionsRoutes.js
```diff
@@ -0,0 +1,83 @@
...
+const errors = require('../../server/errors');
+const AccountType = require('../AccountType');
+const catapult = require('catapult-sdk');
+
+const { address, networkInfo, mosaicRestriction } = catapult.model;
+const { uint64 } = catapult.utils;
+
+module.exports = {
+	register: (server, db, services) => {
+		const accountIdToAddress = (type, accountId) => ((AccountType.publicKey === type)
```
Vektrat: same as metadata, im unsure if this is unique here or if it can be reused, either way, move to `routeUtils` (after merge)

Vektrat:
addresses #74

PR #133, opened 2019-09-26
Upgrade MockServer to allow multiple http methods

Vektrat:
i've tested with mosaicRestrictions and it works - i can update the file when this gets merged


Vektrat:
Addresses #131

PR #171, opened 2019-10-23
Added endpoint for owned mosaics

### LINKED ISSUE #11: Add endpoint for owned mosaics
Similar to how it was possible to *get namespaces from an account*, there should be an endpoint to retrieve **mosaics owned per account**.

Reference: https://nemtech.github.io/api/endpoints.html#operation/getNamespacesFromAccount

martinayora:
Fixes #11

### DIFF
--- a/catapult-sdk/src/plugins/mosaic.js
+++ b/catapult-sdk/src/plugins/mosaic.js
@@ -51,6 +51,15 @@ const mosaicPlugin = {
 			ownerAddress: ModelType.binary,
 			duration: ModelType.uint64
 		});
+
+		builder.addSchema('ownedMosaicDescriptor', {
+			id: ModelType.uint64HexIdentifier,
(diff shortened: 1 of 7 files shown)


catapult-sdk/src/plugins/mosaic.js
```diff
@@ -51,6 +51,15 @@ const mosaicPlugin = {
...
 		});
+
+		builder.addSchema('ownedMosaicDescriptor', {
+			id: ModelType.uint64HexIdentifier,
+			supply: ModelType.uint64,
+			startHeight: ModelType.uint64,
+			ownerPublicKey: ModelType.binary,
+			ownerAddress: ModelType.binary,
+			duration: ModelType.uint64
+		});
```
Vektrat: do we need this? we have `mosaicDescriptor.mosaic` which is exactly the same

PR #177, opened 2019-10-28
Fixed incoming transactions by address

### LINKED ISSUE #170: Querying transactions/incoming by address does not work for accounts not present in the blockchain
related to #164 , creating an issue because it is an specific case that _has_ a solution, while we are unsure we can provide a solution for the rest of transaction endpoints

---
modify the db call to always be queried by address instead of publicKey

martinayora:
Fixes #170.

### DIFF
--- a/rest/src/routes/accountRoutes.js
+++ b/rest/src/routes/accountRoutes.js
@@ -22,6 +22,9 @@ const routeResultTypes = require('./routeResultTypes');
 const routeUtils = require('./routeUtils');
 const AccountType = require('../plugins/AccountType');
 const errors = require('../server/errors');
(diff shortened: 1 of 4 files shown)


rest/src/routes/accountRoutes.js
```diff
@@ -74,5 +78,20 @@ module.exports = {
 						.then(transactionSender.sendArray('accountId', res, next)));
 			});
 		});
+
+		server.get('/account/:accountId/transactions/incoming', (req, res, next) => {
```
Vektrat: i believe we can keep the old logic and just pass an address, we can rename the `accountAddress`/`publicKey` to `account` or similar to generalize

PR #197, opened 2019-11-15
Updated transaction schemas to fushicho2

### LINKED ISSUE #193: Update transaction schemas to fushicho2
https://github.com/nemtech/catbuffer/tree/fushicho2

martinayora:
Fixes #193


catapult-sdk/src/modelBinary/embeddedEntityCodec.js
```diff
@@ -31,8 +31,11 @@ const embeddedEntityCodec = {
 	 */
 	deserialize: parser => {
 		const entity = {};
+		entity.embeddedTransactionHeader_Reserved1 = parser.uint32();
```
evias: This is missing `size: uint32`. EmbeddedTransaction schema inlines `SizePrefixedEntity` to header: https://github.com/nemtech/catbuffer/blob/fushicho2/schemas/transaction.cats#L17

dgarcia360: ``blockHeaderCoded ``and ``transactionsCodec`` does not treat either ``size``.  Perhaps they are filtering it before, but definitely something to double-check before merging.

Vektrat: it's not missing the size

catapult-sdk/src/plugins/lockSecret.js
```diff
@@ -46,10 +46,9 @@ const lockSecretPlugin = {
 		});
 
 		builder.addTransactionSupport(EntityType.secretLock, {
-			mosaicId: ModelType.uint64HexIdentifier,
-			amount: ModelType.uint64,
-			duration: ModelType.uint64,
 			secret: ModelType.binary,
+			mosaic: ModelType.uint64HexIdentifier,
```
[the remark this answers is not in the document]
Vektrat: this schema has not changed

it is true that marti changed the SDK serializer and removed the `amount` due to a confusion and we have reverted this, but the linked comment here refers to a REST schema, which we have still not worked on (starting today) - i believe we cannot test this collection without the SDK because we have no secretlocks in our database, but still, this has not changed in the schemas, so most likely it is fine as it is

martinayora:
> seems wrong to me. `UnresolvedMosaic = UnresolvedMosaicId + Amount`. It is not a simple field name change.

We did not just changed the name, we also added the amount that I mistakenly removed.

Vektrat:
> Open API talks about mosaicId
> 
> https://github.com/nemtech/nem2-openapi/blob/45-f2/spec/plugins/lock_secret/schemas/SecretLockTransactionBodyDTO.yml#L4

we will put mosaicId for consistency then, but i repeat: this was not a f2 change, and has always been like this

PR #201, opened 2019-11-19
Updated db models to fushicho2

### LINKED ISSUE #198: Update db models to comply with fushicho2 changes

martinayora:
Fixes #198


catapult-sdk/src/plugins/aggregate.js
```diff
@@ -89,6 +89,7 @@ const requireCodecs = txCodecs => {
 const aggregatePlugin = {
 	registerSchema: builder => {
 		const aggregateSchema = {
+			transactionHash: ModelType.binary,
```
Vektrat: this has been implemented prematurely, it's a bugfix from a field that is not in core yet but has to be added

PR #218, opened 2019-12-02
Improved fetching account transactions by publicKey and address

### LINKED ISSUE #214: Get transactions by address
Review which transactions can be retrieved by address in addition to public key.

https://github.com/nemtech/catapult-rest/blob/bd4f912fbf98cbfb85ae66e4d0f21ecb993060bd/rest/src/routes/accountRoutes.js#L70-L93

martinayora:
Fixes #214


martinayora:
I have removed old account transactions tests because they where only working by public key and not by address. Then figured out that queryTransactions operation is missing tests in CatapultDb tests.

I have opened a new issue: #219

PR #236, opened 2019-12-18
Task/233 remove page size step

### LINKED ISSUE #233: Remove step parameter from endpoint pagination
this parameter was originally included so that certain endpoints could benefit from a caching layer at REST level, however we will remove it due to the following circumstances:

- it is restricting clients
- there is and there hasn't yet been any caching mechanisms or studies done on this front anyway
- the only endpoint that would probably most benefit from this is the blocks listing one, that is static as it grows in data

note that this parameter is provided at config level and forced at bootstrapping later on

martinayora:
Fixes #233 

Notice the endpoint behavior has slightly changed. Here https://nemtech.github.io/nem2-openapi/#operation/getBlocksByHeightWithLimit it says:

`Block height. If height -1 is not a multiple of the limit provided, the inferior closest multiple + 1 is used instead.`

This used to make sense when limit was rounded to page size with step parameter. Now that we allow any page size this rounding doesn't make sense anymore. So height is not altered in anyway. Except for height 0 when it is changed to 1.

Limit is still sanitized to be in range of page size min and max.

### DIFF
--- a/catapult-sdk/package.json
+++ b/catapult-sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "catapult-sdk",
-  "version": "0.7.20.5",
+  "version": "0.7.20.6",
   "description": "Catapult SDK core",
   "main": "_build/index.js",
   "scripts": {
--- a/rest/package.json
+++ b/rest/package.json
@@ -1,6 +1,6 @@
 {
   "name": "catapult-api-rest",
-  "version": "1.0.20.5",
+  "version": "1.0.20.6",
   "description": "",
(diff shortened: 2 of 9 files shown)


martinayora:
Adding @dgarcia360 as a reviewer to validate the new behaviour.

Vektrat:
holding merge pending on @dgarcia360

Vektrat:
remember to increase version number please

dgarcia360:
The new behavior looks correct, will update the docs 👍

PR #237, opened 2019-12-18
Enable keccak for public networks

### LINKED ISSUE #190: Enable keccak for public networks
@decentraliser

Vektrat:
Fixes #190 

check comment in issue, PR not ready to merge, missing reviews, tests, and potentially changing variable name, if everything else is fine

### DIFF
--- a/catapult-sdk/package.json
+++ b/catapult-sdk/package.json
@@ -3,9 +3,12 @@
   "version": "0.7.20.6",
   "description": "Catapult SDK core",
   "main": "_build/index.js",
+  "config": {
+    "hashingFunction": "sha3"
+  },
   "scripts": {
     "clean": "rimraf _build && mkdir _build && rimraf _build_browser && mkdir _build_browser",
-    "build": "ncp src/ _build && cross-env BABEL_ENV=production babel src -d _build_browser --source-maps",
+    "build": "ncp src/ _build && cross-env BABEL_ENV=production HASHING_FUNCTION=$npm_package_config_hashingFunction babel src -d _build_browser --source-maps",
     "rebuild": "npm run clean && npm run build",
     "test": "mocha --full-trace --recursive",
     "test:coverage": "nyc npm test && nyc report --reporter=text-lcov",
--- a/catapult-sdk/package.json
+++ b/catapult-sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "catapult-sdk",
-  "version": "0.7.20.6",
+  "version": "0.7.20.7",
   "description": "Catapult SDK core",
   "main": "_build/index.js",
   "config": {
--- a/catapult-sdk/src/crypto/sha3Hasher.js
+++ b/catapult-sdk/src/crypto/sha3Hasher.js
@@ -21,9 +21,26 @@
 /** @module crypto/sha3Hasher */
 const arrayUtils = require('../utils/arrayUtils');
 const convert = require('../utils/convert');
-const { sha3_256, sha3_512 } = require('js-sha3');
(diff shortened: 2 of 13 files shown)


ministras:
@Vektrat, rest throws an error while launching:
$ node _build/index.js resources/rest.json
/home/nem/catapult-rest/catapult-sdk/_build/crypto/sha3Hasher.js:40
        throw new Error('A valid hashing function has to be specified in the configuration file <package.json>');

Not sure do I need to explicitly define HASHING_FUNCTION before building?
Or is it enough as is in package.json with "config": { "hashingFunction": "sha3"} ?

p.s. in file "catapult-sdk/test/crypto/sha3Hasher_spec.js" there is 'keccakk' with double "k" is it right?

kaiyzen:
When starting the rest server set HASHING_FUNCTION=keccak as env or in command while starting

The double kk is probably typo unless trying to assert a failure with malformed passed hashing function type

Vektrat:
double kk is a typo that i forgot while doing some testing and forcing wrong names

my idea was that env func shouldnt be manually provided, but i guess since every env is a little different there may be complications, just curious how are you running the server?

ministras:
@kaiyzen You mean setting system wide global like "export HASHING_FUNCTION=keccak"
In catapult-sdk/package.json there is HASHING_FUNCTION=$npm_package_config_hashingFunction as I understand this will override env and assign some $npm_package_config_hashingFunction which should come from npm? Why? Where do we use npm package config?
Can we pass this var to "yarn start"?

@Vektrat what was the idea behind env func? We define "hashingFunction": "sha3" in catapult-sdk/package.json that is not enough as I see. Do I understand correctly that you intended to pass hashing func parameter during every launch of rest server?

Servers set up on Azure VPS with Ubuntu 18.04 - 1 dual node, 1 peer node, 1 api node + rest
"bin/catapult.server ~/catapult/"
"yarn start resources/rest.json"

Vektrat:
> @Vektrat what was the idea behind env func? We define "hashingFunction": "sha3" in catapult-sdk/package.json that is not enough as I see. Do I understand correctly that you intended to pass hashing func parameter during every launch of rest server?

the idea of using `$npm_package_config_hashingFunction` was that it would not be necessary to manually provide `HASHING_FUNCTION` every time the server was started as long as it uses the `_build_browser` output, i asked @kaiyzen if it was working fine in his environments and he affirmed, i assumed he didn't manually provide that variable, but maybe he did? providing it manually still works, obviously, but i'd have preferred for this not to be so explicit in running commands

kaiyzen:
Currently default way things are being delivered is via a build from master, which defaults to sha3 in the package.json config of base, then in the bootstrap tool and couple variants of it we execute the start command passing in keccak for default compat with public-test type deployments.  This allows us to have single build then with various usages just pass in a runtime the intended configured behavior

If it seems like its an issue for folks who aren't using bootstrap, testnet, other related tooling we can revisit to see if multiple builds for each type are required but passing simple param in run time to the container seems rather simple solution to start with

ministras:
Understood guys, thanks for explanation. 
Indeed, it is useful for automation tasks. As I use shell scripts for building separate components it doesn't matter, but for bootstrap and eventually for automation using containers it makes sense. I think there is no need to make multiple builds, just would be good to clarify for others in readme, that in some cases you will need to pass simple param when starting.
Another big question is SHA3 vs Keccak. First I don't see any benefits switching from standard. Secondly nobody mentions any difference in the algorithm that NEM (CAT) uses differently from standard SHA3, because Keccak is the SHA3. That makes NEM especially exotic chain when you deal with cryptography on a hardware level.

PR #238, opened 2019-12-20
Task/t186 fix duplicate connections

### LINKED ISSUE #186: Ensure only one open connection to API node
Referring to this https://github.com/nemtech/catapult-rest/pull/6 as it is still an issue

Original quote:
> When rest gateway receives a burst of transactions on startup, it will open a new connection for each request as connectionService has not saved the connection while waiting for the verification to complete.

Vektrat:
Fixes #186 (#6), and hopefully the random `Address already in use` failure from tests, though it's just a wild guess

basically added a cached "pending for authentication" object to store connections that have not been authenticated yet since this does not happen until `lease()`, in this window of time if multiple requests are received different connections are spawned with conflicting addresses

### DIFF
--- a/rest/src/connection/connectionService.js
+++ b/rest/src/connection/connectionService.js
@@ -39,6 +39,9 @@ module.exports.createConnectionService = (config, connectionFactory, authPromise
 	const node = config.apiNode;
 	const clientKeyPair = createKeyPairFromPrivateKeyString(config.clientPrivateKey);
 	const aliveConnections = {};
+	// the connection is not persisted until authentication is done, and this may take a while,
+	// to avoid duplicate connections those are cached in the following object
+	const authenticatingConnectionPromises = {};
 
 	/**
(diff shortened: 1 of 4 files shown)


rest/src/connection/connectionService.js
```diff
@@ -65,15 +68,31 @@ module.exports.createConnectionService = (config, connectionFactory, authPromise
...
-			}, reject);
+				return serverConnection;
+			}).catch(err => {
+				if (isPersistent)
+					delete authenticatingConnectionPromises[node];
+
+				// reject and additionally, return the connection for possible queued connections on `authenticatingConnectionPromises`
+				reject(err);
+				throw err;
+			});
```
Vektrat: not using `finally` for
```
if (isPersistent)
        delete authenticatingConnectionPromises[node];
```

because the `resolve()` code is executed before the `finally()`s which can lead into conflicts

PR #241, opened 2020-01-13
Task/229 fix transactions incoming aggregate

### LINKED ISSUE #229: /transactions/incoming and aggregate transactions
Currently ``/account/:accountId/transactions/incoming`` does not return  aggregate transactions with at least one inner transaction with recipientAddress matching the :accountId. 

Imagine a situation where someone wants to get the sum of ``cat.currency`` received from multiple accounts. Since some of the TransferTransactions could be wrapped under Aggregate Transactions, the only way to achieve this would be querying and filtering all the confirmed transactions on the client-side.

martinayora:
Fixes #229.


Vektrat:
update version at `package.json` files

PR #251, opened 2020-02-07
Added network fee recommendation endpoint

### LINKED ISSUE #98: Add REST endpoint for network fee recommendation
Currently it is hard to tell what amount of transaction fee must be included for a transaction to be confirmed in reasonable time.

With this issue, it is proposed to add an endpoint to catapult-rest that will permit to retrieve information about network fees from the node.

Averages and medians can be computed over a period of `maxRollbackBlocks` blocks.

Following information can be returned by this endpoint:

- averageFeeMultiplier = average fee multiplier over period of `maxRollbackBlocks`.
- medianFeeMultiplier = median fee multiplier over period of `maxRollbackBlocks`.
- highestFeeMultiplier = highest fee multiplier over period of `maxRollbackBlocks`.
- lowestFeeMultiplier = lowest fee multiplier over period of `maxRollbackBlocks`.

This endpoint could also include information about **effectively paid fees** over the period of `maxDifficultyBlocks` such as:

- Effective rental fee to register/extend a namespace per block.
- Effective rental fee to create a subnamespace.
- Effective rental fee to create a mosaic.

martinayora:
Fixes #98.

### DIFF
--- a/rest/resources/rest.json
+++ b/rest/resources/rest.json
@@ -1,8 +1,7 @@
 {
   "network": {
     "name": "mijinTest",
-    "description": "catapult development network",
-    "maxRollBackBlocks": 300
+    "description": "catapult development network"
   },
 
   "port": 3000,
@@ -74,5 +73,7 @@
       "maxsize": 20971520,
       "maxFiles": 100
     }
-  }
+  },
+
+  "numBlocksTransactionFeeStats": 300
 }
--- a/rest/resources/rest.json
+++ b/rest/resources/rest.json
@@ -1,7 +1,8 @@
 {
   "network": {
     "name": "mijinTest",
-    "description": "catapult development network"
+    "description": "catapult development network",
+    "maxRollBackBlocks": 300
   },
 
   "port": 3000,
--- a/catapult-sdk/package.json
+++ b/catapult-sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "catapult-sdk",
-  "version": "0.7.20.10",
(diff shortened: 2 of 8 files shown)


rest/resources/rest.json
```diff
@@ -1,7 +1,8 @@
 {
   "network": {
     "name": "mijinTest",
-    "description": "catapult development network"
+    "description": "catapult development network",
+    "maxRollBackBlocks": 300
```
dgarcia360: Since the node operator might change this value might be changed and not correspond to the network ``maxRollBackBlocks``, I would suggest moving this property outside ``network`` and call it ``transactionFeesBlocks``.

AnthonyLaw: Just wonder, it should use `maxDifficultyBlocks ` rather than `maxRollBackBlocks` ?

Vektrat: > Since the node operator might change this value might be changed and not correspond to the network `maxRollBackBlocks`, I would suggest moving this property outside `network` and call it `transactionFeesBlocks`.

thought i answered this yesterday but i just see that it was not really posted -_- i agree with moving it outside of network, though i'd try to come up with something different from both of your suggestions that expresses what this seating really does, definitely not `maxRollBackBlocks` because it can be confused with the server setting, `transactionFeesBlocks` sounds better but i'd still add some extra words to make it more self-explanatory, even if a bit long of a setting name

Vektrat: let's use `numBlocksForTransactionFeeStats` and close this

PR #259, opened 2020-02-14
Task/257 hash updates

### LINKED ISSUE #257: Update hashing functions
Gimre:

> Just for the heads up, there are changes coming regarding key generation and addresses, I'll try to update test-vectors later today.
> Short summary is as follows:
> 
> - key generation will always use ed25519 with SHA512 (not sha3/keccak), this is default for many libraries and will hopefully make things easier (although that change was dictated by something completely different)
> - address generation will always use sha3 (no keccak) so in short: `network_byte || ripemd(sha3_256(...)) `, and sha3_256 to get address "checksum"
> 
> ofc side-effect is no backward compatibility with NIS1 addresses

Vektrat:
Fixes #257 

Removed a lot of unused code that was common between the old library and the SDK


Vektrat:
⚠️⚠️⚠️  need to increase version number before merging ⚠️⚠️⚠️

PR #284, opened 2020-03-13
Added block transactions type filter

### LINKED ISSUE #249: Add type filter to /block/<:height>/transactions
The endpoint [getBlockTransactions](https://nemtech.github.io/nem2-openapi/#operation/getBlockTransactions)  does not currently support filtering by transaction type [(example)](https://nemtech.github.io/nem2-openapi/#operation/getAccountConfirmedTransactions).

martinayora:
Fixes #249.

### DIFF
--- a/catapult-sdk/package.json
+++ b/catapult-sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "catapult-sdk",
-  "version": "0.7.20.23",
+  "version": "0.7.20.24",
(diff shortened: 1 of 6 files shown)


Vektrat:
> is this still needed as we are discussing moving all transaction related query to a single endpoint?

this is needed, and since we don't know when the other changes which are still under discussion will be implemented, we are closing pending issues

PR #299, opened 2020-04-07
Wrap connections with tls

Vektrat:
the `monitor` subproject is missing the updates, but i might investigate if it is worth keeping, and either way i think it's better if we deploy the changes here and then worry about the monitor


Vektrat:
@dgarcia360 in case you want to add some mentions to the docs:

- does not affect clients/SDKs
- we have added 3 new options to the config file
- we have removed (i think) 2 options from the config files (client and public key), if those were documented feel free to remove
- wrapped connections with TLS1.3

- Reminder: in a quite recent commit we upgraded REST to be supported only by Node.js 12, instead of 8 and 9

dgarcia360:
@Vektrat Thanks for the notes!

PR #300, opened 2020-04-07
Refactored pagination and transactions

Vektrat:
⚠️ not to be merged before the TLS updates ⚠️


rest/test/db/CatapultDb_spec.js
```diff
@@ -1512,16 +1512,559 @@ describe('catapult db', () => {
...
 
+	describe('queryPagedDocuments 2', () => {
+		const testPublicKey = {
+			one: test.random.publicKey(),
+			two: test.random.publicKey()
+		};
+		const testAddress = {
+			one: keyToAddress(testPublicKey.one),
+			two: keyToAddress(testPublicKey.two)
+		};
```
Vektrat, line 1533:
```
const account1 = {
    publicKey: ...,
    address: ...
}
const account2 = ...
```

Vektrat:
missing tests for `formattingUtils.js` and `formatters.js`

rest/test/routes/transactionRoutes_spec.js
```diff
@@ -18,96 +18,275 @@
...
-				'transactionId',
-				Valid_Transaction_Hashes,
-				Valid_Transaction_Hashes.map(convert.hexToUint8)
-			);
-			addHomogeneousCheck(Valid_Transaction_Hashes, Valid_Object_Ids[0]);
+				testCases.forEach(testCase => {
+					runParseFilterTest(testCase.filter, testCase.param, testCase.value);
+				});
+
+				it('transactionTypes', () => {
```
Vektrat, line 204: why is this test standalone and not with the other filters tests above?

rest/test/routes/transactionRoutes_spec.js
```diff
@@ -18,96 +18,275 @@
...
+						valid: { object: { transactionId: ids[0] }, parsed: [parsedIds[0]], printable: ids[0] },
+						validMultiple: { object: { transactionIds: ids }, parsed: parsedIds },
+						invalid: { object: { transactionId: '12345' }, error: `transactionId ${errorMessage}` },
+						invalidMultiple: {
+							object: { transactionIds: ['12345', ids[0], ids[1]] },
+							error: `element in array transactionIds ${errorMessage}`
+						}
+					},
+					dbApiName,
+					type: 'transactionWithMetadata'
```
Vektrat, line 68: i know this test already existed but this happens in the new test as well:

please **import** and use `routeResultTypes.transaction` instead

rest/src/routes/routeUtils.js
```diff
@@ -128,10 +131,15 @@ const routeUtils = {
 	 * @returns {object} Parsed paging options.
 	 */
 	parsePagingArguments: args => {
-		const parsedOptions = { id: undefined, pageSize: 0 };
```
Vektrat: reverted to old function, those changes make no sense anymore here (probably forgot to delete) - need to change related tests

rest/test/routes/routeUtils_spec.js
```diff
@@ -235,6 +243,173 @@ describe('route utils', () => {
...
+		});
+
+		it('fails when invalid page number is provided', () => {
+			// Act:
+			expect(() => routeUtils.parsePaginationArguments({ pageNumber: '12aa' }, servicesConfigPageSize))
+				.to.throw('pageNumber is not a valid unsigned integer');
+		});
 	});
 
 	describe('sender', () => {
```
Vektrat, line 415: missing `sendPage()` tests

rest/test/routes/routeUtils_spec.js
```diff
@@ -235,6 +243,173 @@ describe('route utils', () => {
...
+			// Assert:
+			expect(options).to.deep.equal({
+				pageSize: 12,
+				pageNumber: 1,
+				sortField: 'id',
+				sortDirection: 1
+			});
+		});
+
+		it('bounds page size to max page size', () => {
```
Vektrat, line 287: binds

rest/test/routes/routeUtils_spec.js
```diff
@@ -235,6 +243,173 @@ describe('route utils', () => {
...
+			// Assert:
+			expect(options).to.deep.equal({
+				pageSize: servicesConfigPageSize.max,
+				pageNumber: 1,
+				sortField: 'id',
+				sortDirection: 1
+			});
+		});
+
+		it('bounds page size to min page size', () => {
```
Vektrat, line 303: binds

rest/test/routes/transactionRoutes_spec.js
```diff
@@ -18,96 +18,275 @@
...
+						params: { state: 'nonsenseSatate' }
+					};
+
+					// Act + Assert
+					expect(() => mockServer.callRoute(route, req)).to.throw('invalid transaction state provided');
+				});
+			});
 		});
 	});
 });
```
Vektrat, line 292: forced `sortField` to `id` now, this must be adapted in the tests, and also have a specific test that shows this behaviour (always defaults sortField to id)

rest/test/routes/transactionRoutes_spec.js
martinayora: Because is the only filter where the filter name and the param are different. That made testCases longer and multiline, making it far less readable. It can be added with the rest of the tests adding one more param to runParseFilterTest, but I felt I wasn't necessary. It can be changed though.

rest/test/db/CatapultDb_spec.js
```diff
@@ -1512,16 +1518,657 @@ describe('catapult db', () => {
...
+				createTransaction(10, [testAddress.one], 1, 0, 0, 0)
+			];
+
+			// Act + Assert:
+			return runDbTest(
+				{ transactions: dbTransactions },
+				db => db.transactions({}, paginationOptions),
+				transactionsPage => {
+					expect(transactionsPage.data.length).to.equal(1);
+					expect(undefined === transactionsPage.data[0].meta.addresses).to.equal(true);
```
Vektrat, line 1873: i think `to.be.undefined` exists, otherwise change to `to.equal(undefined)` just like you did above - review this in the whole document

rest/test/db/CatapultDb_spec.js
```diff
@@ -1512,16 +1518,657 @@ describe('catapult db', () => {
...
+			const dbTransactions = [
+				createTransaction(10, [testAddress.one], 1, 0, 0, 0)
+			];
+
+			// Act + Assert:
+			return runDbTest(
+				{ transactions: dbTransactions },
+				db => db.transactions({}, paginationOptions),
+				transactionsPage => {
+					expect(transactionsPage.data.length).to.equal(1);
```
Vektrat, line 1872: not needed here

rest/test/db/CatapultDb_spec.js
```diff
@@ -1512,16 +1518,657 @@ describe('catapult db', () => {
...
+				runPageNumberTest(1, 10);
+				runPageNumberTest(2, 10);
+				runPageNumberTest(3, 10);
+				runPageNumberTest(4, 2);
+				runPageNumberTest(5, 0);
+			});
+		});
+	});
+
+	describe('transactions', () => {
```
Vektrat, line 1785: missing test without any filters (i think? unsure if i've commented this elsewhere)

Vektrat, line 1785: missing sort test

rest/test/db/CatapultDb_spec.js
```diff
@@ -1512,16 +1518,657 @@ describe('catapult db', () => {
...
+		const testPublicKey = {
+			one: test.random.publicKey(),
+			two: test.random.publicKey(),
+			three: test.random.publicKey()
+		};
+		const testAddress = {
+			one: keyToAddress(testPublicKey.one),
+			two: keyToAddress(testPublicKey.two),
+			three: keyToAddress(testPublicKey.three)
+		};
```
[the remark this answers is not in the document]
martinayora: Here this change would be confusing, as they are reused in multiple tests and multiple times inside the same test.

rest/src/db/CatapultDb.js
```diff
@@ -430,7 +430,8 @@ class CatapultDb {
 			if (filters.height)
 				conditions.push({ 'meta.height': convertToLong(filters.height) });
 
-			conditions.push({ 'meta.aggregateId': { $exists: false } });
+			if (!filters.embedded)
+				conditions.push({ 'meta.aggregateId': { $exists: false } });
```
Vektrat, line 434: the default behaviour is of exclusion, so this condition will always need to be defined, something like:

```
conditions.push({ 'meta.aggregateId': { $exists: true === filters.embedded } });
```

the filter is not defined as mandatory in the params, so undefined needs to be considered as well, the suggested should also work with undefined

to test in `transactions()`:
- that default query without providing embedded, returns only non-embedded results
- that query with embedded=true, returns only embedded
- that query with embedded=false, returns only non-embedded results

i think the previous tests should not be working... because they should have been done with default behaviour (since this filter didn't exist), and with the updated version, there is no default behaviour, but since transcationRoutes is forcing a default to embedded (always exists, even if not provided it will be undefined and it evaluates to false) they pass

if the current code was like `if (false === filters.embedded`), which would have been the more precise way to do it, the old tests would probably fail (because embedded will be undefined) - tests failing is a desired behaviour because it would prove the code is working as expected

PR #329, opened 2020-05-21
Task/t317 review namespaces endpoints

### LINKED ISSUE #317: Review namespaces endpoints
In accordance with nemtech/symbol-openapi#107 and using the new pagination

martinayora:
Fixes #317.


dgarcia360:
@martinayora If it's available on MongoDB should be ok to also add it. Please, let me know if you finally add it to update the OpenAPI accordingly.

Vektrat:
> @martinayora If it's available on MongoDB should be ok to also add it. Please, let me know if you finally add it to update the OpenAPI accordingly.

we will keep the same functionality as before, having it filter implicitly unless a requirement arises in the future

catapult-sdk/src/plugins/namespace.js
```diff
@@ -65,11 +65,19 @@ const namespacePlugin = {
 			namespaces: { type: ModelType.array, schemaName: 'namespaceDescriptor' }
 		});
 
+		builder.addSchema('namespaceDescriptorWithId', {
```
Vektrat, line 68: i'm not sure i understand this...

why do we need both `namespaceDescriptor` and `namespaceDescriptorWithId`? both endpoints using those schemas should return `id`, `meta`, and `namespace`, right? it is possible that `/namespaces/:namespaceId` is not parsing the id right now

related: `/namespaces/:namespaceId` is using `findNamesByIds` which uses `sanitizer.deleteIds`, not sure this is correct

martinayora: You were right, id in the `/namespaces/:namespaceId` endpoint wasn't being parsed. Now this is fixed. Regarding `sanitizer.deleteIds` it was changed to `sanitizer.renameId` in a previous commit.

PR #414, opened 2020-07-17
Height parsed as uint64 in query params

### LINKED ISSUE #324: Treat height as Uint64 in the query params
"blocked" by #296

martinayora:
Fixes #324.

### DIFF
--- a/rest/src/plugins/receipts/receiptsRoutes.js
+++ b/rest/src/plugins/receipts/receiptsRoutes.js
@@ -27,7 +27,7 @@ module.exports = {
 		server.get('/statements/transaction', (req, res, next) => {
 			const { params } = req;
 			const filters = {
-				height: params.height ? routeUtils.parseArgument(params, 'height', 'uint') : undefined,
+				height: params.height ? routeUtils.parseArgument(params, 'height', 'uint64') : undefined,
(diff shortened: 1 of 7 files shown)


martinayora:
@Vektrat  Should the height be parsed with `uint64` instead of `uint64Hex` when used as sort field or offset?

https://github.com/nemtech/catapult-rest/blob/task/T324-query-params-height-as-uint64/rest/src/routes/blockRoutes.js#L46

Vektrat:
> @Vektrat Should the height be parsed with `uint64` instead of `uint64Hex` when used as sort field or offset?
> 
> https://github.com/nemtech/catapult-rest/blob/task/T324-query-params-height-as-uint64/rest/src/routes/blockRoutes.js#L46

if i remember correctly we put hex because openapi defined it like so, not sure if i asked david at that time, @dgarcia360 can you take a look? should we change this to decimal rather than hex?

Vektrat:
> if i remember correctly we put hex because openapi defined it like so, not sure if i asked david at that time, @dgarcia360 can you take a look? should we change this to decimal rather than hex?

nevermind, however i look at it it should be decimal (same as the param), and openapi is not defining the `offset/sort` type strictly because it is dynamic, so change it here, and make sure david see this just in case anyway

PR #425, opened 2020-07-20
Docker deployments using Travis

fboucquez:
Hi guys

In order to improve the CI experience, I've upgraded Travis to deploy docker images in this PR. 

Each `master` build will build and push docker image with a name like `symbolplatform/symbol-rest:1.2.0-alpha` into the docker hub. The idea is that we can automatically pull the newly created image down the line into Bootstrap (local) or a possible  DevNet (aws). Using rest alpha will speed up parallel development, SDKs and clients can test your just finish feature and provide feedback before an official release

Once you are happy with your development, you can `tag` master, then Travis will pick up the version, build and deploy into docker hub from that tag. The deployed image will be `symbolplatform/symbol-rest:1.2.0`. This image can be then be picked by TestNet (aws) or Bootstrap (local) when doing a suite-wide release. 

The image create is based on the current TestNet [DockerFile](https://github.com/Wayonb/symbol-testnet-bootstrap/blob/master/common/dockerfiles/symbol-rest/Dockerfile). The only difference is that it doesn't need to clone the repo,  just copy the Travis working dir with the current master or tagged code. 

Before merging this PR, you need to provide the following ENV in [Travis's settings ](https://travis-ci.com/github/nemtech/catapult-rest/settings)

- DOCKER_IMAGE_NAME. example: `symbolplatform/symbol-rest`
- DOCKER_USERNAME. The docker user with permission to push into the docker orga (symbolplatform)
- DOCKER_PASSWORD. The docker user's password with permission to push into the docker orga (symbolplatform)

@dgarcia360 , what would it be the best NGL orga in dockerhub? Would we keep `symbolplatform`? Let us know when you have the NGL's bot docker info and credentials.

I've tested the process in my fork and it works fine (master and tag build).

Selection_151

The process adds 4 minutes to the master build. Branch or PR builds won't run the last deploy, it doesn't slow down the PR review process.

We can speed up Travis (if needed):
- I don't think `node_modules` folders are being cached in Travis as they are in subfolders. The dependencies are downloaded and compiled each `test` build (this is a current issue)
- Set up Travis to cache docker images and checkpoints (current one is `node:12.18.1-alpine`). This is not out of the box but there are some solutions [posted](https://github.com/travis-ci/travis-ci/issues/5358). 
- Copy `node_modules` from Travis working dir to the docker images when creating (by removing */node_modules from. dockerignore). For this, we need to change the parent image from `node:12.18.1-alpine` to `12.18.2-stretch` or an image compatible with Travis so the already compiled and downloaded `node_modules` folders work on the image by just copying them. This will speed up release but increase the image size from 743MB to 1.14GB (alpine images are smaller). Also `ash` is replaced by `bash` when running the image (bash is used catapult-server and catapult-tools). @Wayonb what do you think? 

NOTES:
- This process is slightly different from other release processes like the SDKs, CLI, open API, etc. There is no release branch that automatically upgrades the version, pushes artifacts, and tags the repo. In this case, you need to manually tag and then upgrade the version in the rest's package.json. I didn't want to change your release process, I just added an automatic deploy step. If you like the other `release` branch process we can upgrade it in future PR.

### DIFF
--- a/docker.sh
+++ b/docker.sh
@@ -1,0 +1,35 @@
+#!/bin/sh
+
+if [ "${DOCKER_IMAGE_NAME}" = "" ]
+then
+  echo "Docker deployment ignored env DOCKER_IMAGE_NAME has not been provided"
+  exit;
+fi
+
+if [ ! "${DOCKER_PASSWORD}" = "" ]
+then
+  echo "Login into docker..."
+  echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+fi
+
+cd rest
+CURRENT_VERSION=$(npm run version --silent)
+cd ..
+
+if [ "$1" = "--release" ]
+then
+   VERSION="${CURRENT_VERSION}"
+else
+   VERSION="${CURRENT_VERSION}-alpha"
+fi
+
+echo "Creating image ${DOCKER_IMAGE_NAME}:${VERSION}"
+docker build -t "${DOCKER_IMAGE_NAME}:${VERSION}" .
+docker images
+
+# Quick test, the expected error is ENOENT: no such file or directory, open '../resources/rest.json'
+echo "Testing image ${DOCKER_IMAGE_NAME}:${VERSION}. NOTE: ENOENT: no such file or directory, open '../resources/rest.json is expected"
+docker run --rm "${DOCKER_IMAGE_NAME}:${VERSION}" npm start
+
+echo "Pushing image ${DOCKER_IMAGE_NAME}:${VERSION}"
+docker push "${DOCKER_IMAGE_NAME}:${VERSION}"
--- a/docker.sh
+++ b/docker.sh
@@ -2,15 +2,24 @@
 
 if [ "${DOCKER_IMAGE_NAME}" = "" ]
 then
-  echo "Docker deployment ignored env DOCKER_IMAGE_NAME has not been provided"
-  exit;
+  echo "Docker deployment error. Env DOCKER_IMAGE_NAME has not been provided"
+  exit 128
 fi
 
-if [ ! "${DOCKER_PASSWORD}" = "" ]
+if [ "${DOCKER_USERNAME}" = "" ]
 then
-  echo "Login into docker..."
-  echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+  echo "Docker deployment error. Env DOCKER_USERNAME has not been provided"
+  exit 128
 fi
+
+if [ "${DOCKER_PASSWORD}" = "" ]
+then
+  echo "Docker deployment error. Env DOCKER_PASSWORD has not been provided"
+  exit 128
+fi
+
+echo "Login into docker..."
+echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
 
 cd rest
 CURRENT_VERSION=$(npm run version --silent)
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,16 @@
+Dockerfile
+.dockerignore
+docker.sh
+
+.idea
+.vscode
+.git
+.gitignore
+.travis.yml
+
+*/node_modules
+*/_build
+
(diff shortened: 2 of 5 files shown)


Vektrat:
looks fine, not too worried about the caching opportunities right now, after all this is not execution time, we can polish with time if it becomes an annoyance

i suppose this is **blocked** until dockerhub account issues are sorted out?

docker.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
+
+if [ "${DOCKER_IMAGE_NAME}" = "" ]
+then
+  echo "Docker deployment ignored env DOCKER_IMAGE_NAME has not been provided"
+  exit;
+fi
+
+if [ ! "${DOCKER_PASSWORD}" = "" ]
```
Vektrat, line 9: why is it checked? there's no else behaviour and if not provided the script just goes on? and `$DOCKER_USERNAME` is not checked?

fboucquez: You can run docker.sh in your dev computer without re-login

Vektrat: yes but what if you run it not in your dev computer without providing the details?

fboucquez: I guess the check could include $DOCKER_USERNAME, $DOCKER_PASSWORD or both

fboucquez: if there are no docker credentials at all (params or pre `docker login`), docker push will fail with an unauthorized error

Vektrat: yes, but after the build - i'd prefer the script to fail if no username and password are provided, after all this is automation, not a script that we would usually expect to be run locally

fboucquez: OK, ignoring, rejecting, or reusing is fine for me.  Let me update to fail if not all params are provided

fboucquez:
are we upgrading master to main? this PR assumes master as it's still the current branch.

Vektrat:
> are we upgrading master to main? this PR assumes master as it's still the current branch.

not sure, maybe, but let's focus on this now and we will see what happens, we can always change it, since i''m not even sure if they want to change repositories yet

docker.sh
fboucquez: Updated...

fboucquez:
https://travis-ci.com/github/nemtech/catapult-rest/settings

I think env variables when `Only available to the master branch` won't be set when building the tag release (deploy may fail when publishing released images). Please don't forget to set DOCKER_IMAGE_NAME too.

Vektrat:
> https://travis-ci.com/github/nemtech/catapult-rest/settings
> 
> I think env variables when `Only available to the master branch` won't be set when building the tag release (deploy may fail when publishing released images). Please don't forget to set DOCKER_IMAGE_NAME too.

yeah i know, i was just wondering why the travis CI was apparently hanging atm... i just can't find any info anywhere

why wouldn't they be set for tag release? tags are in the master branch as well are they not?

fboucquez:
> why wouldn't they be set for tag release? tags are in the master branch as well are they not?

I'm not 100% if master only env will be picked by tag builds. We will see I guess.

Vektrat:
i've been checking this for a few hours and i cannot seem to understand what is wrong with the whole environment (builds not consistent, failures, failing to update to coveralls, etc)

probably the issue is on github/coveralls/travis integrations and the fact that this comes from an other repository branch master, so in the end i think we will merge and fix if there's any issue once within the repository

Vektrat:
ok it worked fine

@fboucquez i'm thinking of replacing the `version-alpha` image with just `catapult-rest-alpha` or something like that, because i don't see the point of having alphas for every version in the repository, i guess it's just used as a "latest" dev version, anything past that seems useless?

fboucquez:
There 3 possible tags from strict to open:

1. x-alpha-timestamp
2. x-alpha
3. alpha/latest

If we just alpha/latest but we fix the server version, a user with an old version of bootstrap may be running a newer version of rest against an old version of the server. I guess it's too volatile. Option 2 seems safer.

Having said that, we can tag all 3 to an image giving the users the flexibility.

https://stackoverflow.com/questions/31963525/is-it-possible-for-image-to-have-multiple-tags

PR #444, opened 2020-07-29
statements/transaction endpoint now accepts multiple receiptTypes

### LINKED ISSUE #440: Allow array filtering in the statements endpoint field `receiptType`
Endpoint: `/statements/transaction`

The filter receiptType it’s possible change to array to accept multiple `receiptType`, just like transaction filter’s transactionTypes

**Reason**: it will easy for client apps to group those data and pagination, because some of receiptType share the same object properties.

Example: receiptType
```
Harvest_Fee
LockHash_Created
LockHash_Completed
LockHash_Expired
LockSecret_Created
LockSecret_Completed
LockSecret_Expired
```
is belong to `balanceChangeReceipt`

So I can easily group those receiptType[] create different filter such as, balanceTransferReceipt, artifactExpiryReceipt and inflationReceipt

martinayora:
Fixes #440.

### DIFF
--- a/rest/src/plugins/receipts/ReceiptsDb.js
+++ b/rest/src/plugins/receipts/ReceiptsDb.js
@@ -55,7 +55,7 @@ class ReceiptsDb {
 			conditions.push({ 'statement.height': convertToLong(filters.height) });
 
 		if (undefined !== filters.receiptType)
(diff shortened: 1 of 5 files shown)


Vektrat:
> Has this change been reflected on the openAPI spec yet?

nope, not yet

PR #445, opened 2020-07-30
New db formatters added

### LINKED ISSUE #437: Create formatters for current pass-through basic types so that everything is explicitly supported
- int
- string
- boolean
- ...?

just create and test the formatters to cover everything that we had until now, no need to put them into use yet

this should take about 2 days max unless there are surprises

martinayora:
Fixes #437.

### DIFF
--- a/rest/src/db/dbFormattingRules.js
+++ b/rest/src/db/dbFormattingRules.js
@@ -20,12 +20,9 @@
 
 const { longToUint64 } = require('./dbUtils');
 const catapult = require('catapult-sdk');
-const { Binary } = require('mongodb');
 
 const { ModelType, status } = catapult.model;
 const { convert, uint64 } = catapult.utils;
-
-const rawUint64ToUint64 = value => (value instanceof Binary ? uint64.fromBytes(value.buffer) : longToUint64(value));
 
 module.exports = {
 	[ModelType.none]: value => value,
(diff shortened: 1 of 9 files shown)


rest/src/db/dbFormattingRules.js
```diff
@@ -35,7 +35,12 @@ module.exports = {
 	[ModelType.statusCode]: value => status.toString(value >>> 0),
 	[ModelType.string]: value => value.toString(),
 	// `uint16` adds support to uint16 arraybuffers in addition to basic uint16
```
Vektrat, line 37: update comment

rest/src/db/dbFormattingRules.js
```diff
@@ -34,8 +34,13 @@ module.exports = {
 	[ModelType.objectId]: value => (undefined === value ? '' : value.toHexString().toUpperCase()),
 	[ModelType.statusCode]: value => status.toString(value >>> 0),
 	[ModelType.string]: value => value.toString(),
-	// `uint16` adds support to uint16 arraybuffers in addition to basic uint16
-	[ModelType.uint16]: value => (value instanceof Binary ? Buffer.from(value.buffer).readInt16LE(0) : value),
+	// `uint` and `int` formatters add support to uint16 arraybuffers in addition to basic uint16
+	[ModelType.uint8]: value => (value instanceof Binary ? Buffer.from(value.buffer).readUInt8(0) : value),
+	[ModelType.uint16]: value => (value instanceof Binary ? Buffer.from(value.buffer).readUInt16LE(0) : value),
```
Vektrat, line 39: not sure why was this originally put... either:
- considered that specific uint16 parsing was required for some reason (though it isn't, i mean, it will display negative values, but just like in every other situation, so i don't see why it should be special in the only case where this formatter is used)
- completeness: this is only used here https://github.com/nemtech/catapult-rest/blob/main/catapult-sdk/src/plugins/restrictions.js#L84, `valueType` _must_ be defined because this is not a schema, but an options object that defines an schema, so it being undefined would be wrong, but we could as well have used `ModelType.none` i believe...

with this, consider removing it (if this is the way to go for the chosen approach)

PR #448, opened 2020-08-05
Strict schema parsers

### LINKED ISSUE #439: Upgrade the schema parser to be strict rather than pass-through
Blocked by #438 

upgrade the schema parser so that only defined fields are forwarded, and dropped otherwise

could be easy but trouble expected, probably one week to implement/test

Vektrat:
Closes #439


Vektrat:
pending for wayon to test (he executed tests but failed because he had to update the java SDK, pending for repetition

fboucquez:
Do I need to update the java SDK?

Vektrat:
> Do I need to update the java SDK?

no

martinayora (approved):

PR #449, opened 2020-08-05
Enhance ws responses wih topic

### LINKED ISSUE #309: Notification from unconfirmedRemoved, partialRemoved and cosignature will be received by all registered address with the same web socket connection
1. Expected behavior and actual behavior.
If multiple accounts register multiple accounts to listen to cosignature using one listener. They should only get notify for transactions that have their address.
2. Steps to reproduce the problem.
Register multiple accounts to listen to cosignature using one listener.

Does not filter using the user address.

```
    /**
     * Returns an observable stream of {@link CosignatureSignedTransaction} for specific address.
     * Each time a cosigner signs a transaction the address initialized,
     * it emits a new message with the cosignatory signed transaction in the even stream.
     *
     * @param address address we listen when a cosignatory is added to some transaction address sent
     * @return an observable stream of {@link CosignatureSignedTransaction}
     */
    public cosignatureAdded(address: Address): Observable<CosignatureSignedTransaction> {
        this.subscribeTo(`cosignature/${address.plain()}`);
        return this.messageSubject.asObservable().pipe(
            filter((_) => _.channelName.toUpperCase() === ListenerChannelName.cosignature.toUpperCase()),
            filter((_) => _.message instanceof CosignatureSignedTransaction),
            map((_) => _.message as CosignatureSignedTransaction),
        );
    }
```

Basically when multiple subscriptions set up in one listener object, without knowing the uri / channel name (e.g. consignatureAdd/{address}) multiple notifications will be received. Discussed internally, and would like rest to have a look.

Ref https://github.com/nemtech/symbol-sdk-typescript-javascript/issues/542

Vektrat:
Closes #309

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ All notable changes to this project will be documented in this file.
 
 The changelog format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [v1.2.1] - 21-Aug-2020
+### Changed
+- Removed `topic` (address) from transactionStatus WS responses.
+- Removed `channelName` from WS transaction metadata.
+- Wrapped WS responses so that the `topic` the client subscribed to is also returned.
+
 ## [v1.2.0] - 6-Aug-2020
 ### Added
 - TLS installation notes.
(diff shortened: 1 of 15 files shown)


fboucquez:
Web socket looks really nice. Great!. I've done the Java one, it's ready to be merged. TS will follow soon. 

I'm keen on merging rest and the sdk, but before doing, is it possible to upgrade the rest [version](https://github.com/nemtech/catapult-rest/blob/main/rest/package.json#L3) (1.2.1?). So we can get a free alpha image for our bootstraps.

@Vektrat @martinayora @rg911

PR #465, opened 2020-09-01
Merged chain height and score endpoints into /chain/info and added finalization data to it

### LINKED ISSUE #461: Upgrade /chain/height to include finalization information
as discussed here https://github.com/nemtech/catapult-rest/issues/457 we will be adding the `latestFinalizedBlock` info to this endpoint

the endpoint should be renamed to `/chain/info` to accommodate the new struct, and with this generalization we suggest that `/chain/score` also merges up into it

spec: https://github.com/nemtech/symbol-openapi/pull/196

Vektrat:
Closes #461


Vektrat:
missing implementation, missing tests

rest/src/routes/chainRoutes.js
```diff
@@ -20,25 +20,20 @@
...
-
-		server.get('/chain/score', (req, res, next) =>
-			db.chainStatisticCurrent().then(curryStripProperties(['height'], currySend(db, res, next))));
+		server.get('/chain/info', (req, res, next) =>
+			Promise.all([
+				db.chainStatisticCurrent(),
+				db.latestFinalizedBlock()
+			]).then(dbResults => {
+				const chainInfoResult = dbResults[0];
+				chainInfoResult.latestFinalizedBlock = dbResults[1].block;
```
fboucquez, line 31: dbResults[1] could be undefined when starting the network. latestFinalizedBlock should be optional here and in the API. 
Also, once there is a value, 

finalizedHeight is not returned, I guess we can remove it from the open API as it's the same value than latestFinalizedBlock.height right?

Jaguar0625: > dbResults[1] could be undefined when starting the network. latestFinalizedBlock should be optional here and in the API.
server will from now on always return something (i.e. the nemesis finalized block information), so this doesn't need to be changed in REST.

martinayora (approved):

PR #479, opened 2020-09-14
Bumped project version

Vektrat:
@segfaultxavi just in case you are interested in this


segfaultxavi:
> @segfaultxavi just in case you are interested in this

From what David told me, I feel like I should be doing something about this...

Is it related to https://docs.symbolplatform.com/compatibility.html ?

Vektrat:
> > @segfaultxavi just in case you are interested in this
> 
> From what David told me, I feel like I should be doing something about this...
> 
> Is it related to https://docs.symbolplatform.com/compatibility.html ?

yes it is, you'll need to update the compatible versions once the team has agreed on freezing them/releasing as a whole

PR #495, opened 2020-10-02
Updated schemas and parsers for unsigned values

### LINKED ISSUE #323: Consider displaying API numbers originating from mongodb as unsigned
currently, unsigned data fetched from mongodb is displayed as signed since [mongodb does not support](https://docs.mongodb.com/manual/reference/bson-types/) unsigned data

consider treating every number in the project as unsigned unless specified since the vast majority are unsigned, or add and parse every number in the schemas even if basic type (which wouldn't be a bad idea anyway from a documentation/explicitness point of view)

related: https://github.com/nemtech/catapult-rest/issues/144

martinayora:
Fixes #323.


Vektrat (approved):
`finProof.version`, `messageGroup.stage` not sure those are uint32 (maybe uint8), but it's not a big deal and not simple to check, so looks fine

rest/src/db/dbFormattingRules.js
```diff
@@ -40,9 +40,10 @@ module.exports = {
 	[ModelType.objectId]: value => (undefined === value ? '' : value.toHexString().toUpperCase()),
 	[ModelType.statusCode]: value => status.toString(value >>> 0),
 	[ModelType.string]: value => value.toString(),
-	[ModelType.uint]: value => convert.int32ToUint32(value),
+	[ModelType.uint8]: value => convert.int32ToUint32(value),
```
Vektrat, line 43: why use `int32ToUint32`? in the `ModelType.uint16` we are using different a different approach

rest/test/db/dbFormattingRules_spec.js
```diff
@@ -94,25 +94,12 @@ describe('db formatting rules', () => {
...
 
-	describe('can format uint type', () => {
-		const testCases = [
-			{ name: 'uint value 0 (min)', value: new Int32(0), formated: 0 },
-			{ name: 'uint8 value 255', value: new Int32(255), formated: 255 },
-			{ name: 'uint16 value 65535', value: new Int32(65535), formated: 65535 },
-			{ name: 'uint32 value 4294967295 (max)', value: new Int32(-1), formated: 4294967295 },
-			{ name: 'uint32 value 2147483647', value: new Int32(2147483647), formated: 2147483647 },
-			{ name: 'uint32 value 2147483648', value: new Int32(-2147483648), formated: 2147483648 }
-		];
```
Vektrat, line 105: why did you drop some of those test cases?

rest/test/db/dbFormattingRules_spec.js
```diff
@@ -94,25 +94,12 @@ describe('db formatting rules', () => {
...
-
-		testCases.forEach(testCase => {
-			it(testCase.name, () => {
-				// Arrange + Act:
-				const result = formattingRules[ModelType.uint](testCase.value);
+	it('can format uint8 type', () => {
```
Vektrat, line 97: probably missing more test cases? not sure if this represents all the possible conversion issues (ie. max uint8 value is 255)

Vektrat (changes requested):
check comments, not sure if changes are needed

rest/src/db/dbFormattingRules.js
martinayora: Both `uint8` and `uint16` maximum values fit inside positive part of `int32`. So we can use `int32ToUint32`. Working on `uint16` case.

martinayora: `int32ToUint32` in fact is not needed because of reason above. Both `uint8` and `uint16` fit inside `int32` positive part. Database will return them in the appropiate format.

rest/test/db/dbFormattingRules_spec.js
martinayora: Test have been recovered.

Vektrat (approved):

PR #518, opened 2020-11-06
Added unlocked_account endpoint

### LINKED ISSUE #516: To create an endpoint to pass unlocked account details from peer node.
this is to get delegated harvesting registration status from the peer node, which is required by wallets

rg911:
Fixed #516

### DIFF
--- a/rest/src/routes/nodeRoutes.js
+++ b/rest/src/routes/nodeRoutes.js
@@ -135,5 +135,22 @@ module.exports = {
 					next();
 				});
 		});
+
+		server.get('/node/unlockedaccount', (req, res, next) => {
+			const { convert } = catapult.utils;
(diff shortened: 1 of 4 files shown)


rest/src/routes/nodeRoutes.js
```diff
@@ -127,11 +159,36 @@ module.exports = {
...
+				.then(connection => connection.pushPull(packetBuffer, timeout))
+				.then(packet => {
+					res.send(
+						buildResponse(packet, nodeTimeCodec, routeResultTypes.nodeTime)
+					);
+					next();
+				});
+		});
+
+		server.get('/node/unlockedaccount', (req, res, next) => {
```
Wayonb, line 177: Do you need the node name? The current implementation will only query the API node.

PR #519, opened 2020-11-07
State /merkle and required new endpoints

fboucquez:
WIP of the /merkle endpoints.

There are several changes on Rest (to be discussed):

1) I have to add get by id/ids of the component hash or the missing natural keys. 
2) I've added empty param searches to be consistent between the different entities (some of the current searches required a param like an address which is not really required if you want to paginate through all entities). 
3) In this PR for now, Rest is returning a raw hex string from API, next step would be to convert this raw hex into a tree array path or also a hash of that so the user can quickly check the value it wants.

The paths for each entity are (account example):

- Paginated search: `GET /accounts`
- Get By Id: `GET /accounts/{address}`
- Get By Ids: `PUT /accounts` (BODY: {addresses:....})
- Gey merkle by Id: `GET /accounts/{address}/merkle`

This multiplied by each entity type (replace any of the above `accounts` with the the entity type):

- `/accounts` (id: encoded/plain address, decoded/hex address)
- `/mosaics` (id: mosaic id hex)
- `/namespaces` (id: namespace id hex)
- `/lock/hash` (id: hash)
- `/lock/secret` (id: composite hash)
- `/metadata` (id: composite hash)
- `/restrictions/account` (id: encoded/plain address, decoded/hex address)
- `/restrictions/mosaic` (id: composite hash)

Multisig could keep the current pattern:

`/account/:address/multisig/merkle`  (note the singular `/account` path)

Or follow the new pattern and allow users doing multsig paginated searches:

- Paginated search: `GET /multisigs`
- Get By Id: `GET /multisigs/{address}`
- Get By Ids: `PUT /multisigs` (BODY: {addresses:....})
- Gey merkle by Id: `GET /multisigs/{address}/merkle`

The idea of the new endpoints is to be consistent and allow users doing reports on a paginated and filtrable list of entities

An example of the `/merkle` payload, I will add the list of paths and probably the hash of the path once I can parse the "raw" hex

image


fboucquez:
Hi @rg911 , we should have a conversation about this to know if this change goes too far.

fboucquez:
One question, atm some entities are allowed can be queried by single (GET) and multiple ids (POST), some only by single id (GET). I've normalized the affected ones in this PR as I needed more by ids enpodints.

rg911 (approved):

PR #523, opened 2020-11-12
Excluding expired namespaces from endpoints

fboucquez:
Fixes https://github.com/nemtech/catapult-rest/issues/515

Expired namespaces (not active or end <= current) are excluded from the rest endpoints.

Currently, expired mosaics are returned in the rest endpoints. Should it be consistent?

Should rest return expired namespaces so the user can check if one of its namespaces has expired or not (instead of just disappear?)


fboucquez:
From today's tech call, expired namespaces will be removed from rest endpoints. Expired mosaics won't be and users would need to check if they are expired based on the `duration` and the `startHeight`

rg911 (approved):

PR #525, opened 2020-11-13
Added node publicKey endpoint

### LINKED ISSUE #524: Extract node's signing key from tls handshake
Node signing key (public) will be used for delegated harvesting node keylink transaction. At the moment the node info returns only the SSL (ca) keys

rg911:
- Added a new endpoint to retrieve node's publicKey from `node.key.pem` cert 
- So sdk can query to get the node key for the key link in delegated harvesting

Fixed #524


rg911:
Closed as node publickey is moved to node/info endpoint in another pr now

PR #526, opened 2020-11-14
Node Public Key exposed in /node/info

fboucquez:
This PR is an alternative to https://github.com/nemtech/catapult-rest/pull/525

Instead of creating a new endpoint, the node public key is added to the node info. 

image

ideally, this PR is temporary until server api provides the node public keys for local and known remote peers.

https://github.com/nemtech/catapult-server/issues/110

If core could return that, node/info and node/peers (for the remote known peers) would work out of the box without rest loading the key from the pem file. Clients would know th node keys of all the known peers via one rest.

### DIFF
--- a/catapult-sdk/src/model/ModelSchemaBuilder.js
+++ b/catapult-sdk/src/model/ModelSchemaBuilder.js
@@ -228,7 +228,8 @@ class ModelSchemaBuilder {
 				friendlyName: ModelType.string,
 				host: ModelType.string,
 				publicKey: ModelType.binary,
-				networkGenerationHashSeed: ModelType.binary
+				networkGenerationHashSeed: ModelType.binary,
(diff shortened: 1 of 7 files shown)


rg911 (approved):

segfaultxavi:
Sorry but `publicKey` and `nodePublicKey` is a horribly confusing choice of words. One should not be part of the other.
Ideally, both should have a different prefix indicating their purpose. 
I don't know this purpose but maybe we can use `signing`, `creation`, `harvesting`, `communication`, `comm`, `tls`...

fboucquez:
Yes @segfaultxavi I agree. 

The old publicKey is not enough, we need to add a new one without breaking too much. 

Please have a look at https://github.com/nemtech/symbol-openapi/pull/244 and we comment from there.

PR #532, opened 2020-11-19
Catapult 10.0.0.4 update

fboucquez:
Added new block types
Updated namespace query to use 'latest'. Added active flag back in memory comparing current height with endHeight.


catapult-sdk/src/model/ModelSchemaBuilder.js
```diff
@@ -58,7 +58,12 @@ class ModelSchemaBuilder {
 				receiptsHash: ModelType.binary,
 				stateHash: ModelType.binary,
 				beneficiaryAddress: ModelType.binary,
-				feeMultiplier: ModelType.uint32
+				feeMultiplier: ModelType.uint32,
+				// optional. How to create subclasses?
+				votingEligibleAccountsCount: ModelType.uint32,
```
rg911, line 63: should it be mapped to a new model schema e.g. `ImportanceBlockFooter`?

rg911: also, if split the footer, should we bring in the block types into the EntityType?

fboucquez: It seems like this manual sdk is only used for known how to parse db fields into payloads. Afaik, It's not used for binary payload serialization and deserialization. 

ideally, this sdk would have all the logic like block type, footer, etc. This logic is already in the catbuffer generated ts. 

In the future, I would like to simplify rest by not using the sdk when moving from mongo db to payload. Just using old good typed mapping without all the formatter/model maps which are hard to follow and untyped. I would only use catbuffer (the generated ts) when we need to serialize and deserialize payloads, communicate with server api, or if we need to generate hashes for proof.

fboucquez: If rest uses the open api generated code (a rest generation version of it), the target payload types would be generated so the mapping would be hard to miss from db to payload

rg911 (dismissed):

rg911 (approved):

PR #542, opened 2020-12-03
Updated version and CHANGELOG.md

fboucquez:


segfaultxavi (approved):
No problems with the text.

CHANGELOG.md
```diff
@@ -3,6 +3,24 @@ All notable changes to this project will be documented in this file.
 
 The changelog format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [v2.2.0] - 3-Dec-2020
+### Added
+- Catapult-server finality (0.10.0.4) support.
```
rg911, line 8: Catapult-server finality (0.10.0.4 - testnet/v3) would be clearer

version.txt
```diff
@@ -1 +1 @@
-2.1.1
+2.2.0
```
rg911, line 1: should we use version suffix to differentiate between main & testnet?
e.g. 2.2.0-tesetnet

rg911 (commented):
we should make a

version.txt
fboucquez: I wouldn't add the testnet prefix/suffix. Server and libs are not adding it.

PR #550, opened 2020-12-13
- Use fork height from network properties

rg911:


rest/src/routes/finalizationRoutes.js
```diff
@@ -35,18 +38,31 @@ module.exports = {
 		const { connections } = services;
 		const { timeout } = services.config.apiNode;
 
+		const readAndParseNetworkPropertiesFile = () => {
+			const readFile = util.promisify(fs.readFile);
+			try {
```
fboucquez, line 43: Is it possible to move the read of configuration files to a service? 

1) be reused here and in the network routes
2) having the object cached in memory, for some reason the file is being read every request.

rest/src/routes/finalizationRoutes.js
```diff
@@ -35,18 +38,31 @@ module.exports = {
...
+				return readFile(services.config.network.propertiesFilePath, 'utf8')
+					.then(fileData => ini.parse(fileData));
+			} catch (err) {
+				// If API node is not updated, then use static fork height
+				return new Promise(resolve => resolve({ fork_heights: { votingKeyLinkV2: '215500' } }));
```
fboucquez, line 48: about the hardcoded value, is this for testnet node not updated? In the future, votingKeyLinkV2 will not be provided for testnet/publicnet with the "main" server

fboucquez (approved):

rest/src/routes/finalizationRoutes.js
rg911: yeah, this is only for those synced with outdated server code.

rest/src/routes/finalizationRoutes.js
rg911: thought about it, but this is a simple temporal patch for testnet/v3 only. won't probably need it once mainnet launched.

fboucquez: Ok, add a comment "to be removed after main merge" or similar.

PR #553, opened 2021-01-16
Fixed aggregate transaction codex issue + 2.3.1 release

rg911:


fboucquez:
This pr and the previous pr remove the transaction multiversion support. We would need to add it back when there is a new transaction version in the future. Hope we have a nice typescript version of it by then.

fboucquez (approved):

PR #554, opened 2021-01-20
Cosigner notification fix

### LINKED ISSUE #278: Cosignatories dont see partial transaction they need to cosign
Expected behaviour:

- given a multisig account A, configure to have 2 minApprovals out of 2 cosignatories B and C
- B signs an aggregate bonded transaction containing a transfer issued by A.
- A, B and C should **all** see the partial transactions 

Actual behaviour:

Currently only A and B are able to see the partial transaction when querying /account/A/transactions/partial and /account/B/transactions/partial. 

**C is not able to see to partial transaction even though he is supposed to cosign.**

Environment:

REST v0.7.20

rg911:
- Added extra step to check if ws subscriber is a cosigner or a multisig, then subscribe the multisig account to the channel for consigning notification
- Added the same check in transaction search endpoint to return transactions assigned to multisig account for all cosigners

Fixed #278


rest/src/db/CatapultDb.js
```diff
@@ -374,6 +375,26 @@ class CatapultDb {
 		const sortConditions = { [sortingOptions[options.sortField]]: options.sortDirection };
 		const conditions = buildConditions();
 
+		// Check multisig graph if address is used in search criteria for cosigning,
+		// Then, show transactions for other cosigers.
+		if (filters.address !== undefined) {
```
fboucquez, line 380: I would move this condition to:

`const buildAccountConditions = () => {`

rest/src/db/CatapultDb.js
```diff
@@ -374,6 +375,26 @@ class CatapultDb {
...
+					const addresses = [];
+					if (multisigEntries.length) {
+						if (multisigEntries[0].multisig.multisigAddresses.length)
+							addresses.push(...multisigEntries[0].multisig.multisigAddresses);
+					}
+					return addresses;
+				}).then(multisigAddresses => {
+					if (multisigAddresses && multisigAddresses.length) {
+						const buffers = multisigAddresses.map(address => address.buffer);
+						conditions['meta.addresses'] = { $in: buffers };
```
fboucquez, line 392: This may be overriding the return { 'meta.addresses': Buffer.from(filters.address) }; above.

why not converting this function to async-await? So there is only one `this.queryPagedDocuments(` exit point

fboucquez (changes requested):

rest/src/db/CatapultDb.js
```diff
@@ -304,24 +305,31 @@ class CatapultDb {
...
+	async transactions(group, filters, options) {
 		const sortingOptions = { id: '_id' };
 
-		const buildAccountConditions = () => {
-			if (undefined !== filters.address)
-				return { 'meta.addresses': Buffer.from(filters.address) };
+		const buildAccountConditions = async () => {
+			// Check multisig graph if address is used in search criteria for cosigning,
+			// Then, show transactions for other cosigers.
+			if (undefined !== filters.address) {
```
Wayonb, line 314: Whats the perf overhead of making this call for non-multisig accounts?

rest/src/server/SubscriptionManager.js
```diff
@@ -93,6 +105,26 @@ class SubscriptionManager {
...
 	}
+
+	/**
+	 * Check if the current subscriber is a cosigner of a multisig account.
+	 * Then subscribe the multisig account for cosigning notification
+	 * @param {string} encodedAddress encloded address of the current subscriber
+	 * @param {object} channel current subscribing channel
+	 * @param {object} client current subscribeing client
+	 */
+	subscribeMultisigAccount(encodedAddress, channel, client) {
```
Wayonb, line 116: What happens when a cosigner is removed from multisig account? They will still receive notifications from the multisig account until the websocket is close?

Wayonb (changes requested):

rest/src/plugins/multisig/multisigUtils.js
```diff
@@ -0,0 +1,77 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Catapult.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+const multisigUtils = {
+	getMultisigGrahp: (db, address) => {
```
fboucquez, line 23: Typo in getMultisigGrahp

rest/src/db/CatapultDb.js
fboucquez: I think it's fine. the extra db call is only done when the address filter is provided and the accountAddress field is indexed in multisigs

rest/src/plugins/receipts/ReceiptsDb.js
```diff
@@ -72,6 +72,17 @@ class ReceiptsDb {
 			conditions[[`statement.receipts.${artifactIdType}`]] = convertToLong(filters.artifactId);
 		}
 
+		if (undefined !== filters.fromHeight || undefined !== filters.toHeight) {
```
fboucquez, line 75: Is this from another PR?

rest/src/db/CatapultDb.js
```diff
@@ -304,24 +305,31 @@ class CatapultDb {
...
+				const multisigEntries = await new MultisigDb(this).multisigsByAddresses([filters.address]);
+				const buffers = [];
 
+				if (multisigEntries.length && multisigEntries[0].multisig.multisigAddresses.length)	{
+					buffers.push(...multisigEntries[0].multisig.multisigAddresses).map(address => address.buffer);
```
fboucquez, line 319: change with:

const buffers = multisigEntries[0].multisig.multisigAddresses.map(address => address.buffer);

Current line has the parentesis in the wrong place

rest/src/db/CatapultDb.js
rg911: How do you the account is multisig or not without checking the db?

rest/src/plugins/receipts/ReceiptsDb.js
rg911: it is

rest/src/server/SubscriptionManager.js
rg911: I don't think it is possible / easy to unsubscribe if cosigner removed from multisig. 
tbh, I don't quite like the idea to hack the rest to subscribe to multisig in ws.  It should be the client to subscribe the multisig account for notification.

fboucquez (approved):

PR #571, opened 2021-02-07
Removed babel from private catapult-sdk

fboucquez:
Removed necessary build (that just copies from src to _build)
People should use the public/regular symbol-sdk for any sdk related op

There are many libs that can be upgraded but this could take a little bit to fully test. This PR just simplifies the build and dev process by removing unnecessary/unused steps. 

Related https://github.com/nemtech/catapult-rest/issues/545


fboucquez:
@Jaguar0625 , would you like to remove the babel build from rest's catapult sdk? As I understand, babel is only to transpile javascript for all browsers. This catapult sdk should not be used for app development as it's internal to the catapult-rest node app (no web). Also, there is no npm package even if somebody wants to include it for some reason.  People should use the old or new sdks anyway.

Removing babel removes lots of dependencies

Jaguar0625 (approved):

fboucquez:
Hey @Jaguar0625. I have rebased this branch. No change more than that. Could you review again?

Jaguar0625 (approved):

PR #577, opened 2021-02-16
Account search endpoint page number issue

### LINKED ISSUE #576: Page number in account search endpoint not changing
Reported by @AnthonyLaw 

The page number in the account search endpoint won't change, it stays at page : 1 even the search criteria specified different pages

rg911:
Fixed #576

### DIFF
--- a/rest/src/db/CatapultDb.js
+++ b/rest/src/db/CatapultDb.js
@@ -266,12 +266,13 @@ class CatapultDb {
 	 * @param {string} collectionName Name of the collection to be queried.
 	 * @param {object} options Pagination options, must contain `pageSize` and `pageNumber` (starting at 1).
 	 * @param {function} mapper to transform each element of the page.
+	 * @param {number} optionalPageIndex optional page index for second query pass
 	 * @returns {Promise.<object>} Page result, contains the attributes `data` with the actual results, and `paging` with pagination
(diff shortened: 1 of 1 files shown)


rest/src/db/CatapultDb.js
```diff
@@ -491,8 +492,7 @@ class CatapultDb {
 					const newConditions = { _id: { $in: accountIds } };
 					// repeat the response with the found and sorted account ids, so that the result can be complete with all the mosaics
 					// Second query set pageIndex to 0;
-					options.pageNumber = 1;
-					return this.queryPagedDocuments(newConditions, [], {}, 'accounts', options)
+					return this.queryPagedDocuments(newConditions, [], {}, 'accounts', options, undefined, 0)
```
fboucquez, line 495: instead of creating a new param, send new arguments as options. 

return this.queryPagedDocuments(newConditions, [], {}, 'accounts', {pageSize, pageNumber:1})

the original options object is not changed and we don't need to add an extra argument.

rg911: Don't think it will work mate, the pageNumber in the pagination obj should return the original number in the query string

fboucquez: Yesp, needs this too:

 fullAccountsPage.pagination.pageNumber = options.pageNumber;

fboucquez (approved):

PR #583, opened 2021-03-01
Zmq socket fix

rg911:
- Only one zmq subscriber socket opened 
- Updated ws route registration. 
- Combined event handlers


gimer:
@rg911 have you tested the change when api node is *syncing* ? (and subscriptions active?)

PR #606, opened 2021-04-27
 circulating supply endpoints

### LINKED ISSUE #605: Network currency current circulating supply
Network currency current circulating supply should be `total supply - uncirculated`

uncirculated tokens:
- unclaimed tokens (on going post-launch opt-in)
- core funding tokens
- other finance team management accounts tokens

rg911:
Fixed #605 
- Added static unclaimed token accounts and core funding accounts list (mainnet)
- Added new endpoint `network/currency/supply/circulating` (totalSupply - unclaimed&coreFunds)
- Renamed the 2 current currency supply endpoints to `network/currency/supply/total` and `network/currency/supply/max`


rest/src/constants/unCirculatedAccounts.js
```diff
@@ -0,0 +1,65 @@
...
+ * Catapult is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Catapult.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+module.exports = [
```
Wayonb, line 22: Should this read from a config file?

rest/src/routes/accountRoutes.js
```diff
@@ -80,5 +84,35 @@ module.exports = {
...
+
+		// CMC specific endpoint
+		const readAndParseNetworkPropertiesFile = () => {
+			const readFile = util.promisify(fs.readFile);
+			return readFile(services.config.apiNode.networkPropertyFilePath, 'utf8')
+				.then(fileData => ini.parse(fileData));
+		};
+
+		// CMC specific endpoint
+		server.get('/network/currency/supply/circulating', (req, res, next) => readAndParseNetworkPropertiesFile()
```
Wayonb, line 96: This endpoint should only be visible in mainnet? For private and testnet, seems to just returned nothing.

Wayonb (changes requested):

rest/src/routes/accountRoutes.js
rg911: Hmm, tested it on private net, works, but the unclaimed total is 0 which is expected, will try again

rest/src/constants/unCirculatedAccounts.js
rg911: So putting it in code is to avoid user configuring it when releasing the full version, since it will be built into the image, since they are ngl accounts, don't really want them to be touched? but guess we can move it to a Json file,

Wayonb (approved):

Issue #604, opened 2021-04-23
REST should return address in plain format rather than hex format. 


AnthonyLaw:
Example:

`"beneficiaryAddress": "6862F5B548DCB21163C530B8CF0FEA2A2C227278B11C5C8F"` (hex format)

`"beneficiaryAddress": "NBRPLNKI3SZBCY6FGC4M6D7KFIWCE4TYWEOFZDY"` (plain format)

rg911:
absolutely agree

fboucquez:
I agree too. Some addresses are aliases so it would be the namespace Id/hex

fboucquez:
Related

https://github.com/nemtech/symbol-openapi/issues/182

zachamo:
Fully agreed.. Worked around this in Python with:

import base64

def hexToAddress(hexKey):
    return base64.b32encode(bytes.fromhex(hexKey)).decode("utf-8").replace("=","")

def addresstoHex(address):
    address.replace("-","")+"="
    return bytes.hex(base64.b32decode(address))

PR #611, opened 2021-05-06
Task/604 address formatting

AnthonyLaw:
### Fix
- added `encodedAddress` in modelType.
- added hexToUint8Reverse (use for NamespaceId) in convert class.
- added `bufferToResolvedAddress` function to convert buffer to Base32 Address and NamespaceId

```json
// Address
"recipientAddress":"TDYGHCCOUL5RB6XOBWN466ZQRGS2FFYVTOCBRNI"

// NamespaceID
"recipientAddress": "BA25CF1CD00FABCA",
```
Todo: 
- [x] beneficiaryAddress
- [x] address
- [x] accountAddress
- [x] ownerAddress
- [x] senderAddress
- [x] recipientAddress
- [x] targetAddress
- [x] multisigAddresses
- [x] cosignatoryAddresses
- [x] addresses
- [x] addressAdditions
- [x] addressDeletions

Resolved: #604


mm-s:
what is the rationale for replacing binary with encoded format?
Thanks

AnthonyLaw:
> what is the rationale for replacing binary with encoded format?
> Thanks

To convert the address field to encoded format is easy for a developer to view data directly into REST return.
 
 Example:
- 68488A227444FC58A7E9740461071B478DBE52720C3E77A6 -> current REST Return.
- NBEIUITUIT6FRJ7JOQCGCBY3I6G34UTSBQ7HPJQ -> encoded format

mm-s:
"directly into REST return"

Since the field name has been reused, how does the change ensure not breaking existing REST service clients? How about external apps that potentially we don't even know they exist?

rg911:
This seems to be a 'breaking' change to all the client apps. But worth having It imo. We Will discuss with other teams see how we can handle the impact

gimer:
@mm-s the problem is that this "binary" isn't actually binary, it's hex-encoded which makes it longer than usual base32 we use to represent as strings...

but yeah, if it's breaking change, shouldn't there be "old" / "new" APIs with "old" being deprecated (and removed in some future update?)

fboucquez:
The SDKs will handle the impact for clients using the SDKs. All our apps are using SDKs. SDK hides how an unresolved address is mapped to an Address/Alias Object. The mappers will create the objects from friendly addresses and alias Ids instead of the hex values.

AnthonyLaw:
I think this PR definitely needs to synchronize with the SDKs upgrades, I don't think it's urgent to merge.

Do we need to support `hex-encoded address` and `base32 address` on API in the future?

If Yes, then will add `encodeAddress` flag in req params, to allow changing the address format.
Else it worth for wait for the SDK upgrade.

fboucquez:
Related PRs
https://github.com/symbol/symbol-openapi/pull/272
https://github.com/symbol/symbol-sdk-typescript-javascript/pull/804

fboucquez:
I think receipts' resolutions are missed....

https://github.com/AnthonyLaw/catapult-rest/blob/task/604_address-formatting/catapult-sdk/src/plugins/receipts.js#L53
https://github.com/AnthonyLaw/catapult-rest/blob/task/604_address-formatting/catapult-sdk/src/plugins/receipts.js#L70

rest/src/db/dbUtils.js
```diff
@@ -83,12 +83,27 @@ const dbUtils = {
...
-	 * Convert Address to Base32 format.
-	 * @param {MongoDb.Binary} addressBinary Address from MongoDb.
-	 * @returns {string} Address in Base32 format.
+	 * Convert binary to resolved address
+	 * @param {MongoDb.Binary} binary Address|NamespaceId from MongoDb.
+	 * @returns {string} AddressBase32|NamespaceId
 	 */
-	bufferToAddressBase32: addressBinary => {
-		const hexToUint8 = convert.hexToUint8(addressBinary.toString('hex'));
+	bufferToResolvedAddress: binary => {
```
fboucquez, line 90: the name would be `bufferToUnresolvedAddress`. The function is returning an Encoded Address or a Mosaic Id Hex.

fboucquez:
One possible issue to consider in the native rest SSL is that this rest only supports one port, so it may not be the best fit for the migration process while apps and wallets are migrating from port HTTP 3000 to HTTPS 3001

rest/src/db/dbUtils.js
```diff
@@ -77,6 +80,31 @@ const dbUtils = {
...
+			// only 8 bytes are relevant to resolve the NamespaceId
+			const namespaceId = hex.substr(2, 16);
+
+			// retun as namespace Id
+			return convert.uint8ToHex(convert.hexToUint8Reverse(namespaceId));
+		}
+
+		// return as Address base 32
+		const hexToUint8 = convert.hexToUint8(hex);
+		return address.addressToString(hexToUint8);
```
Jaguar0625, line 107: why this instead of always addressToString? that is how we normally represent aliases, i think?

Jaguar0625: cc @gimer

rest/src/db/dbFormattingRules.js
```diff
@@ -49,5 +49,6 @@ module.exports = {
 	// `uint64HexIdentifier` requires branching accountRestrictions->restrictionAdditions provides uint64 as binary
 	[ModelType.uint64HexIdentifier]: value => uint64.toHex(value instanceof Binary ? uint64.fromBytes(value.buffer) : longToUint64(value)),
 	[ModelType.int]: value => value.valueOf(),
-	[ModelType.boolean]: value => value
+	[ModelType.boolean]: value => value,
+	[ModelType.encodedAddress]: value => bufferToUnresolvedAddress(value)
```
Jaguar0625, line 53: you need to update messageFormattingRules too or WS will break

Jaguar0625 (changes requested):

rest/src/db/dbUtils.js
AnthonyLaw: for my understanding, `addressToString` method is only way convert Uint8 to an encoded address.

Jaguar0625: i meant returning alias as namespace id seems strange to me. i don't think we use that representation anywhere else, so i'd _just_ use `addressToString`

rest/src/db/dbUtils.js
```diff
@@ -77,6 +80,31 @@ const dbUtils = {
 			return { [sortFieldDbRelation[options.sortField]]: { [1 === options.sortDirection ? '$gt' : '$lt']: offset } };
 		}
 		return undefined;
+	},
+
+	/**
+	 * Convert binary to resolved address
```
Jaguar0625, line 86: unresolved

rest/src/server/messageFormattingRules.js
```diff
@@ -35,5 +36,6 @@ module.exports = {
 	[ModelType.uint64]: value => uint64.toString(value),
 	[ModelType.uint64HexIdentifier]: value => uint64.toHex(value),
 	[ModelType.int]: value => value,
-	[ModelType.boolean]: value => value
+	[ModelType.boolean]: value => value,
+	[ModelType.encodedAddress]: value => bufferToUnresolvedAddress(value)
```
Jaguar0625, line 40: please add tests

rest/test/db/dbUtils_spec.js
```diff
@@ -149,4 +149,39 @@ describe('db utils', () => {
...
+			// Act:
+			const result = dbUtils.bufferToUnresolvedAddress(object);
+
+			// Assert:
+			expect(result).to.equal('BA25CF1CD00FABCA');
+		});
+
+		it('cannot convert from invaild data type.', () => {
+			// Arrange
+			const object = '99caab0fd01ccf25ba000000000000000000000000000000';
```
Jaguar0625, line 178: please use capital hex letters

catapult-sdk/src/utils/convert.js
```diff
@@ -213,6 +213,22 @@ const convert = {
 			throw Error(`input '${input}' is out of range`);
 
 		return (input & 0xFFFFFFFF) >>> 0;
+	},
+
+	/**
+     * Reversed convertion hex string to a uint8 array.
+     * @param {string} input A hex encoded string.
+     * @returns {Uint8Array} A uint8 array corresponding to the input.
+     */
```
Jaguar0625, line 222: check indent

catapult-sdk/src/utils/convert.js
```diff
@@ -213,6 +213,22 @@ const convert = {
...
+	hexToUint8Reverse: input => {
```
Jaguar0625, line 223: `hexToUint8(input).reverse()` (i'm actually not sure if you need a separate function for this)

rest/test/db/dbUtils_spec.js
```diff
@@ -149,4 +149,39 @@ describe('db utils', () => {
...
+			const object = new Binary(Buffer.from('99caab0fd01ccf25ba000000000000000000000000000000', 'hex'));
+
+			// Act:
+			const result = dbUtils.bufferToUnresolvedAddress(object);
+
+			// Assert:
+			expect(result).to.equal('BA25CF1CD00FABCA');
+		});
+
+		it('cannot convert from invaild data type.', () => {
```
Jaguar0625, line 176: 1. invalid
2. no periods in test descriptions

catapult-sdk/src/utils/convert.js
AnthonyLaw: Yeah, I agree, actually the idea is come from the old sdk.
I will remove it 👍🏼

rest/src/db/dbUtils.js
AnthonyLaw: I got what you mean now, the reason to add namespace id  is to handle `Alias address`.

This is the simple. 

image

http://ngl-dual-501.testnet.symboldev.network:3000/transactions/confirmed/3D81C7E22AE0504ABD86EBBBF998E8A1F1F9E6D4C329523ED31760D93B8735C0

Jaguar0625: @gimer - what are your thoughts on this?

Jaguar0625 (dismissed):

gimer (dismissed):

fboucquez:
Approved, we would need to review and merge these ones too

Related PRs
symbol/symbol-openapi#272
symbol/symbol-sdk-typescript-javascript#804

fboucquez (approved):

rest/src/db/dbUtils.js
gimer: ugh, I misread the code this is _not_ ok, this method returns two completely different objects :/

fboucquez: What about if Rest returns either the encoded address or the namespace id hex (like in other places) when working on unresolved addresses?

```
"recipientAddress": "NATNE7Q5BITMUTRRN6IB4I7FLSDRDWZA35C4KNQ", <-real address
"recipientAddress": "291F837E059AE13C", <-Alias
```

Related PR:

https://github.com/symbol/symbol-openapi/pull/272/files#diff-299dfd070f15158c760575d45de8aeb702e4820df3540fbecbe34bcc15f88695R3

AnthonyLaw: What is the expected result, if that is namespace Id hex on `recipientAddress`?

or we should not allow namespace id appear on "recipientAddress"?

Jaguar0625: i like rg911 suggestion to return object with encoded address and optional alias namespace id. or even just always returning encoded address might be sufficient.

i think it would be confusing for consumers to return either encoded address or namespace id.

AnthonyLaw: I believe our goal is to solve the confusion problem for the API.
I will support always returning encoded addresses for the address field will be great.

We need another PR for this.
 - Always resolve the `alias` to `address`

Jaguar0625 (changes requested):
would prefer to change the alias formatting before merging

Jaguar0625 (approved):

Jaguar0625 (approved):

fboucquez:
@AnthonyLaw do you have examples of what is this PR rendering for unresolved address fields? What rest returns when `recipientAddress` is a real address? What rest returns when the `recipientAddress` is an alias?

AnthonyLaw:
@fboucquez sorry for the late reply.

> `recipientAddress` is a real address? 
Yes.

98E0D138EAF2AC342C015FF0B631EC3622E8AFFA04BFCC56 -> TDQNCOHK6KWDILABL7YLMMPMGYRORL72AS74YVQ

> `recipientAddress` is an alias?
No, but this is what we need to solve: https://github.com/symbol/catapult-rest/issues/652

To reduce the confusion, we should only return the Address, rather than the alias.

rest/src/db/dbUtils.js
```diff
@@ -77,6 +79,18 @@ const dbUtils = {
...
 		return undefined;
+	},
+
+	/**
+	 * Convert binary to Unresolved address
+	 * @param {MongoDb.Binary} binary Address|NamespaceId from MongoDb.
+	 * @returns {string} AddressBase32|NamespaceId
+	 */
+	bufferToUnresolvedAddress: binary => {
+		if (!(binary instanceof MongoDb.Binary))
```
fboucquez, line 90: What happens if the binary is an alias? Like 0x99C1ED94FF2D65C0AF000000000000000000000000000000 ?

fboucquez (changes requested):
I have tested this PR and it doesn't render aliases correctly.

All the "addresses" we are formating here are actually unresolved addresses, where the value can either be a real address or a namespace id. 

If you try this PR against testnet, you will see this:

image

http://localhost:3000/transactions/confirmed?height=141765

The same happens with all the unresolved addresses

image

The namespace id format is incorrect, it's not either the current hex format or a namespace id

When the "address" field is actually an alias, we should either render the binary hex or the namespace friendly id as explained before, at least in this pr. Resolving the alias could be done in another PR, but this pr should not break the alias rendering. Be aware that resolving the alias would require db access, it's a feature change.

Note: to try testnet:

got to ./rest and run 
`yarn bootstrap-start-testnet`
and then in another terminal
`yarn start:dev`

Wait for testnet to be synced a little bit, another (lower) example of this issue is:

http://localhost:3000/transactions/confirmed?height=846

You can pull more examples of blocks with aliases 

http://localhost:3000/statements/resolutions/address?order=asc

Note 2: I've broken the merge, could you fix the `dbUtils.js` @AnthonyLaw ? I think I cannot push because you are using a fork.

Jaguar0625 (approved):

fboucquez:
@Jaguar0625 , do you think that rendering aliases as an invalid address is ok (please see the screenshots above with all the AAAAs)?

rest/src/db/dbUtils.js
fboucquez: Catbuffer allows aliases in the (unresolved)address fields and mongo stores it. Rest needs to render that, AAAA doesn't seem right to me...

Jaguar0625:
yes, that is a valid unresolved address. i'm not sure i see the problem?

gimre-xymcity:
> @Jaguar0625 , do you think that rendering aliases as an invalid address is ok? Please see the screenshots above with all the AAAAs

I'm not sure what you're saying, this technically _is_ a valid address... (not valid in the sense of IsValidAddress) (that is exactly how it looks in transaction itself)

fboucquez:
> I'm not sure what you're saying, this technically _is_ a valid address... (not valid in the sense of IsValidAddress)
> (that is exactly how it looks in transaction itself)

Yes, I'm talking in the sense of IsValidAddress

This is the spec for the related open API https://github.com/symbol/symbol-openapi/pull/272

https://github.com/symbol/symbol-openapi/blob/4939747f84089bbc640e772e63787ed2405f11f2/spec/schemas/UnresolvedAddress.yml

You can see that the OR links to a NamespaceId.yml which is a user-friendly namespace identifier 16 hex format

https://github.com/symbol/symbol-openapi/blob/4939747f84089bbc640e772e63787ed2405f11f2/spec/schemas/NamespaceId.yml

like: 85BBEA6CC462B244, a 16 chars hex.

I think the AAAA style address could be confusing and it cannot be used as a query param like the 16 hex. Examples:

http://ngl-dual-101.testnet.symboldev.network:3000/namespaces (16 hex id)
http://ngl-dual-101.testnet.symboldev.network:3000/namespaces/EE0EDBB5737CA68C (16 hex param)

https://github.com/symbol/symbol-openapi/blob/main/spec/schemas/NamespaceId.yml
https://github.com/symbol/symbol-openapi/blob/main/spec/plugins/namespace/routes/getNamespaceMerkle.yml#L7
https://github.com/symbol/symbol-openapi/blob/main/spec/plugins/namespace/routes/getNamespace.yml#L7
https://github.com/symbol/symbol-openapi/blob/main/spec/parameters/query/targetId.yml

One benefit of this PR is that you can use the real address from the payload as a query param, ideally, the namespace id can be used as a query param too (at least until we figure out the alias resolution).

If you think the AAAA addresses are fine, ok to merge.

Jaguar0625:
i think this PR is fine to merge. longer term we might want to consider @rg911 suggestion: https://github.com/symbol/catapult-rest/pull/611#discussion_r686777918

fboucquez (approved):

PR #624, opened 2021-05-17
fix: median fees 0 in network/fees

### LINKED ISSUE #621: Median fee multipler should not be 0
Median fee multiplier should use the defaultMultipler if the calculated blocks are all empty. (to prevent 0 value). it should be used for any block with a multiplier of zero

rg911:
fix: median fees 0 in network/fees
    
  Fixed #621
  
  - Replaced 0 fees (empty block) by defaultDynamicFeeMultiplier in transaction fees calculation
  - applied above changes for averageFeeMultiplier and medianFeeMultiplier
  - updated unit tests

### DIFF
--- a/rest/src/routes/networkRoutes.js
+++ b/rest/src/routes/networkRoutes.js
@@ -48,6 +48,8 @@ module.exports = {
 				.then(fileData => ini.parse(fileData));
 		};
 
+		const sanitizeInput = value => value.replace(/[^0-9]/g, '');
+
 		server.get('/network', (req, res, next) => {
 			res.send({ name: services.config.network.name, description: services.config.network.description });
 			next();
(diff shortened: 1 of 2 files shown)


fboucquez:
There is something odd with the commit history. Changes already in dev seem to be here to? Like the CMC ones?

fboucquez (changes requested):
This branch needs to rebased so it only shows the changes related to the fix. CMC ones have been merged to the dev branch already

rest/src/routes/networkRoutes.js
```diff
@@ -69,10 +71,16 @@ module.exports = {
 		server.get('/network/fees/transaction', (req, res, next) => {
 			const numBlocksTransactionFeeStats = services.config.numBlocksTransactionFeeStats || 1;
 			const latestBlocksFeeMultiplier = db.latestBlocksFeeMultiplier(numBlocksTransactionFeeStats);
-			return Promise.all([readAndParseNodePropertiesFile(), latestBlocksFeeMultiplier]).then(feeMultipliers => {
```
fboucquez, line 72: Configuration files should be loaded and sanitized once, not every request. Each request could just use the loaded object. 
This is a refactoring that's better to have types, so I would refactor/reimplement this once we have Typescript

rest/src/routes/networkRoutes.js
```diff
@@ -69,10 +71,16 @@ module.exports = {
...
+				), 10);
+				const defaultedFeeMultipliers = feeMultipliers[1].map(f => (0 === f ? defaultDynamicFeeMultiplier : f));
 				res.send({
-					averageFeeMultiplier: Math.floor(average(feeMultipliers[1])),
-					medianFeeMultiplier: Math.floor(median(feeMultipliers[1])),
+					averageFeeMultiplier: Math.floor(average(defaultedFeeMultipliers)),
+					medianFeeMultiplier: Math.floor(median(defaultedFeeMultipliers)),
 					highestFeeMultiplier: Math.max(...feeMultipliers[1]),
 					lowestFeeMultiplier: Math.min(...feeMultipliers[1]),
 					minFeeMultiplier: Number(feeMultipliers[0].node.minFeeMultiplier.replace('\'', ''))
```
fboucquez, line 86: should it be sanitizeInput(feeMultipliers[0].node.minFeeMultiplier)

rg911: don't think so, it is already number

fboucquez (approved):

PR #641, opened 2021-07-29
feat: Native HTTPS support

yilmazbahadir:
(Drafted for contribution purposes, when agreed on the defaults Readme will be updated and the PR will be ready for merge)
- Added support for native HTTPS
    - When the `protocol: HTTPS` in `rest.json` application will check if the  `sslKeyPath` and `sslCertificatePath` properties are set in the `rest.json` otherwise will throw an error and won't start the application.
    - Default protocol setting in `rest.json` is `HTTPS` (if `protocol` value is undefined it'll default to `HTTPS` as well), on the other hand, on `symbol-bootstrap` configuration(`shared.yml`), it defaults to `HTTP` instead; so that the user can choose which way they want to configure (Native(default) | Automatic | None).


rest/src/server/bootstrapper.js
```diff
@@ -93,23 +94,59 @@ module.exports = {
 
 	/**
 	 * Creates a REST api server.
-	 * @param {array} crossDomainConfig Configuration related to access control, contains allowed host and HTTP methods.
+	 * @param {array} config Application configuration
```
fboucquez, line 97: {object} instead of {array}?

rest/src/server/bootstrapper.js
```diff
@@ -93,23 +94,59 @@ module.exports = {
...
+			serverOptions = { ...serverOptions, key: sslKey, certificate: sslCertificate };
+		}
+		// create the server using a custom formatter
+		const server = restify.createServer(serverOptions);
 
 		// only allow application/json
 		server.pre(catapultRestifyPlugins.body());
 
+		// crossDomainConfig Configuration related to access control, contains allowed host and HTTP methods.
+		const crossDomainConfig = config && config.crossDomain ? config.crossDomain : undefined;
```
fboucquez, line 148: `const crossDomainConfig = config && config.crossDomain || {}`
better?

fboucquez:
Looks pretty good!

rest/src/server/bootstrapper.js
```diff
@@ -93,23 +94,59 @@ module.exports = {
...
-		});
+		};
+		if (!config.protocol)
+			winston.warn('Protocol is not configured explicitly in the configuration, defaulting to HTTPS.');
+
+		const protocol = config.protocol ? config.protocol : 'HTTPS';
+		winston.warn(`Using protocol: ${protocol}`);
+		// create http or https server based on the configuration and default will be https
+		if (config && 'HTTPS' === protocol) {
+			let sslKey; let
```
fboucquez, line 116: avoid `let` if possible...

rest/resources/rest.json
```diff
@@ -5,6 +5,9 @@
   },
 
   "port": 3000,
+  "protocol": "HTTPS",
+  "sslKeyPath": "",
+  "sslCertificatePath": "",
```
Jaguar0625, line 10: question: are we using a separate certificate for REST => client as opposed to the one already used to communicate with the server?

cc: @gimer

rest/src/server/bootstrapper.js
Jaguar0625: seems like all the other param description end with '.'

rest/src/server/bootstrapper.js
```diff
@@ -93,23 +94,59 @@ module.exports = {
...
+	createServer: (config, formatters, throttlingConfig) => {
+		let serverOptions = {
 			name: '', // disable server header in response
 			formatters: {
 				'application/json': formatters.json
 			}
-		});
+		};
+		if (!config.protocol)
+			winston.warn('Protocol is not configured explicitly in the configuration, defaulting to HTTPS.');
```
Jaguar0625, line 110: isn't the default in the config HTTPS? so, the user would be explicitly turning it off by leaving it empty?

in any event, it might be better to make this a noisy error (i.e. fail to start) instead of allowing a potential misconfiguration.

rest/src/server/bootstrapper.js
```diff
@@ -93,23 +94,59 @@ module.exports = {
...
+
+		const protocol = config.protocol ? config.protocol : 'HTTPS';
```
Jaguar0625, line 112: fyi, since this is JS, this can be written more concisely as `config.protocol || 'HTTPS'`

rest/src/server/bootstrapper.js
```diff
@@ -93,23 +94,59 @@ module.exports = {
...
+		winston.warn(`Using protocol: ${protocol}`);
```
Jaguar0625, line 113: shouldn't this be info, not a warning?

warnings and errors often lead to alerts from monitoring software

rest/src/server/bootstrapper.js
```diff
@@ -93,23 +94,59 @@ module.exports = {
...
+		// create http or https server based on the configuration and default will be https
+		if (config && 'HTTPS' === protocol) {
```
Jaguar0625, line 115: you've already accessed config above so, config check is redundant

rest/src/server/bootstrapper.js
```diff
@@ -93,23 +94,59 @@ module.exports = {
...
+				sslCertificate;
+			if (!config.sslKeyPath) {
+				throw new Error('Server default is HTTPS but no SSL Key found, '
+				+ '\'sslKeyPath\' property in the configuration must be provided.');
+			} else {
+				try {
+					sslKey = fs.readFileSync(config.sslKeyPath);
+				} catch (err) {
+					throw new Error(`SSL Key file cannot be found at the path: ${config.sslKeyPath}`);
+				}
```
Jaguar0625, line 126: this catch does not seem to add much value. since it is throwing a new error distinct from the original one, the original callstack is being lost, which might make diagnosing problems harder. [e.g. maybe the root cause is a file perm issue]

rest/src/server/bootstrapper.js
```diff
@@ -93,23 +94,59 @@ module.exports = {
...
+		const protocol = config.protocol ? config.protocol : 'HTTPS';
+		winston.warn(`Using protocol: ${protocol}`);
+		// create http or https server based on the configuration and default will be https
+		if (config && 'HTTPS' === protocol) {
+			let sslKey; let
+				sslCertificate;
+			if (!config.sslKeyPath) {
+				throw new Error('Server default is HTTPS but no SSL Key found, '
+				+ '\'sslKeyPath\' property in the configuration must be provided.');
+			} else {
```
Jaguar0625, line 121: since the if case alwayse exited the function (via a throw), the else is redundant.

rest/src/server/bootstrapper.js
```diff
@@ -93,23 +94,59 @@ module.exports = {
...
+				try {
+					sslKey = fs.readFileSync(config.sslKeyPath);
+				} catch (err) {
+					throw new Error(`SSL Key file cannot be found at the path: ${config.sslKeyPath}`);
+				}
+			}
+			if (!config.sslCertificatePath) {
+				throw new Error('Server default is HTTPS but no SSL Certificate found,'
```
Jaguar0625, line 129: since this and previous string are identical except var name, consider refactoring by using a template string

rest/test/server/bootstrapper_spec.js
```diff
@@ -751,6 +754,41 @@ describe('server (bootstrapper)', () => {
...
+		it('starts https and throws error when the protocol is not defined', done => {
+			expect(() => createServer({ port: 3001 })).to.throw();
+			done();
+		});
+
+		it('starts http when the protocol is HTTP', done => {
+			createServer({ port: 3001, protocol: 'HTTP' });
+			done();
+		});
+	});
```
Jaguar0625, line 790: are you missing a success test?

yilmazbahadir:
Thanks for the feedbacks guys! Refactored and pushed. Although the idea to draft the PR was to revisit after agreeing on the defaults.

rest/src/server/bootstrapper.js
yilmazbahadir: Users can only turn it off by setting it to HTTP, leaving empty will default to the HTTPS and in that case, the REST server won't start since it'll be asking for the SSL key and certificate, so yes, it'll be a noisy error.

rest/resources/rest.json
yilmazbahadir: Manually installed/managed REST applications(outside of symbol-bootstrap) will default to HTTPS, meaning the user has to have a valid DNS record and a valid SSL Certificate issued by a Trusted Authority, has to configure the key and cert in the rest.json and he will be maintaining his certificates and domain records himself. That's what we agreed on the Discord AFAIK.

rest/test/server/bootstrapper_spec.js
Jaguar0625: ?

yilmazbahadir: Yes if I understood correctly; I've tried to have a HTTPS server routes test(s) earlier so that I can verify it's responding expectedly when on HTTPS; I couldn't find a way to overcome the "self-signed" certificate issue, this was one of the reasons the PR was in draft for now. But now I'm working on finding a way to overcome that issue.

rest/src/server/bootstrapper.js
yilmazbahadir: already fixed.

rest/test/server/bootstrapper_spec.js
yilmazbahadir: I found a workaround for hippie(a request wrapper used in the project) to call an HTTPS endpoint(doesn't work with hippie(httpsApp) unfortunately), created a test case to make sure it's responding expectedly, pushed, please take a look, it'd be nice to have a test case for `wss` as well.

rest/resources/rest.json
gimer: @Jaguar0625 y, it's a separate one

rest/src/server/bootstrapper.js
```diff
@@ -57,7 +58,8 @@ const createCrossDomainHeaderAdder = crossDomainConfig => (req, res) => {
 		if ('*' !== crossDomainResponseHeaders['Access-Control-Allow-Origin'])
 			crossDomainResponseHeaders.Vary = 'Origin';
 
-		Object.keys(crossDomainResponseHeaders).forEach(header => res.header(header, crossDomainResponseHeaders[header]));
+		Object.keys(crossDomainResponseHeaders).forEach(header =>
```
gimer, line 61: ?

rest/src/server/bootstrapper.js
```diff
@@ -88,44 +90,95 @@ const catapultRestifyPlugins = {
...
-	 */
-	createServer: (crossDomainConfig, formatters, throttlingConfig) => {
-		// create the server using a custom formatter
-		const server = restify.createServer({
+   * Creates a REST api server.
+   * @param {object} config Application configuration (see rest.json).
+   * @param {object} formatters Formatters to use for formatting responses.
+   * @param {object} throttlingConfig Throttling configuration parameters, if not provided throttling won't be enabled.
+   * @returns {object} Server.
+   */
```
gimer, line 121: changed tabs to spaces?

rest/test/server/bootstrapper_spec.js
```diff
@@ -751,6 +754,70 @@ describe('server (bootstrapper)', () => {
...
+		});
+
+		it('starts http when the protocol is HTTP', done => {
+			createServer({ port: 3001, protocol: 'HTTP' });
+			done();
+		});
+
+		it('handles HTTPS routes successfully', done => {
+			// to suppress self signed certificate errors
+			process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
```
gimer, line 793: imo, fine for testing purposes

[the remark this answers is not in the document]
Jaguar0625: assuming this is to avoid a client (not server) error, this is ok. but please clarify in the comment.

rest/src/server/bootstrapper.js
fboucquez: It looks like a code style change that is lint compatible. If we set up prettier in rest, prettier will unify any of those changes.

rest/test/server/bootstrapper_spec.js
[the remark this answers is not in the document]
fboucquez: Yes, it's for the rest client in the unit tests to call the HTTPS rest server using self-signed certificates.

fboucquez:
@Jaguar0625 @gimer I am continuing @yilmazbahadir work while he is on holiday. Fixed style and feedback. Could you review again?

rest/src/server/bootstrapper.js
fboucquez: Reverted code style change

rest/src/server/bootstrapper.js
fboucquez: Fixed

rest/test/server/bootstrapper_spec.js
```diff
@@ -751,6 +754,70 @@ describe('server (bootstrapper)', () => {
...
+				.to.throw('No SSL Key found, \'sslKeyPath\' property in the configuration must be provided.');
+			done();
+		});
+
+		it('throws error when the certificate path is missing', done => {
+			expect(() => createServer({ port: 3001, protocol: 'HTTPS', sslKeyPath: `${__dirname}/certs/restSSL.key` }))
+				.to.throw('No SSL Certificate found, '
+				+ '\'sslCertificatePath\' property in the configuration must be provided.');
+			done();
+		});
```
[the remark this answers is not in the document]
fboucquez: Could you clarify? the unit tests seem fine.

fboucquez:
One possible issue to consider in the native rest SSL is that this rest only supports one port, so it may not be the best fit for the migration process while apps and wallets are migrating from port HTTP 3000 to HTTPS 3001. Eventually we would close HTTP/3000 so it won't be a problem then

fboucquez (approved):

gimer (approved):

Jaguar0625 (approved):

Issue #636, opened 2021-07-23
please include block timestamp in receipt data returned by `statements/transaction`


Jaguar0625:
problem:
i want to retrieve receipts and correlate them with real world time, but receipt payload only contains block height.

workaround:
for each receipt, query block to extract timestamp. this can then be used with epoch to calculate good enough time. unfortunately for X receipts, this results in X+1 HTTP requests.

suggestion:
Perform the queries on the REST side and include timestamp in each receipt.

Issue #637, opened 2021-07-23
please include block timestamp and feeMultiplier in transaction returned by transactions/confirmed


Jaguar0625:
similar to  #636 but for transactions/confirmed endpoint

fboucquez:
Is it possible to add the size to the transactions collection meta? It would be nice to be able to know that information and the effective fee (size*feeMultipler) without serializing the transaction (especially aggregate transactions where the inner transactions are stored in other entries).

I know this could require a data migration, but just a point to consider for the future.

Jaguar0625:
size is already returned by rest afaik

fboucquez:
Uff, yes, you are right. Please ignore my comment.

fboucquez:
@Jaguar0625 , should inner transactions include blocks' timestamp and feeMultiplier in the meta too?

![image](https://user-images.githubusercontent.com/5390558/131697345-a2a829c6-eb21-4094-9007-acd4e600a3bb.png)

Jaguar0625:
yes, i would think so

PR #647, opened 2021-08-16
Timestamp in recipient

fboucquez:
Fixes https://github.com/symbol/catapult-rest/issues/636

It adds timestamp next to the id in transaction statements.

image

This implementation runs a second query looking for blocks for the given block heights. It does not use aggregate queries. I tried but failed to create an aggregate query between the `transactionStatements` and `blocks` collections.

If a mongo aggregate query is desired, I can try again to make it work. Note that the aggregate is between nested properties, not top-level properties.


fboucquez:
Once this is approved, the solution can be replicated for https://github.com/symbol/catapult-rest/issues/637

catapult-sdk/test/plugins/receipts_spec.js
```diff
@@ -59,7 +59,7 @@ describe('receipts plugin', () => {
 			// - mosaicResolutionStatement
 			// - transactionStatement
 			['addressResolution', 'mosaicResolution', 'transaction'].forEach(statementType => {
-				expect(Object.keys(modelSchema[`${statementType}Statement`]).length).to.equal(2);
+				expect(Object.keys(modelSchema[`${statementType}Statement`]).length).to.equal(3);
 				expect(modelSchema[`${statementType}Statement`]).to.contain.all.keys(['statement']);
```
Jaguar0625, line 63: you might want to update this check

rest/test/plugins/receipts/ReceiptsDb_spec.js
```diff
@@ -181,12 +185,12 @@ describe('receipts db', () => {
...
 					createTransactionStatement(10, 100, [createReceipt({ recipientAddress: testAddress1 })]),
-					createTransactionStatement(20, 110, [
+					createTransactionStatement(20, 100, [
 						createReceipt({ recipientAddress: testAddress1 }),
 						createReceipt({ recipientAddress: testAddress2 })
 					]),
-					createTransactionStatement(30, 120, [createReceipt({ recipientAddress: testAddress2 })]),
-					createTransactionStatement(40, 130, [])
+					createTransactionStatement(30, 100, [createReceipt({ recipientAddress: testAddress2 })]),
+					createTransactionStatement(40, 100, [])
```
Jaguar0625, line 193: why did you need to change these?

rest/test/db/utils/dbTestUtils.js
```diff
@@ -269,6 +269,17 @@ const runDbTest = (dbEntities, collectionName, createDbFacade, issueDbCommand, a
 		.then(() => db.close());
 };
 
+const insertItems = (dbEntities, collectionName) => {
```
Jaguar0625, line 272: i think insertEntities would be more consistent with other naming in this file

rest/src/db/CatapultDb.js
```diff
@@ -225,6 +225,17 @@ class CatapultDb {
...
 	}
 
+	blocksAtHeights(heights) {
+		if (!heights.length)
+			return Promise.resolve([]);
+
+		return this.queryDocumentsAndCopyIds(
+			'blocks',
+			{ 'block.height': { $in: heights } },
+			{ projection: { 'meta.transactionMerkleTree': 0, 'meta.statementMerkleTree': 0 } }
```
Jaguar0625, line 235: why this projection?

rest/test/db/CatapultDb_spec.js
```diff
@@ -443,6 +443,42 @@ describe('catapult db', () => {
 				block => expect(block).to.deep.equal(stripExtraneousBlockInformation(renameId(seedBlock)))
 			);
 		});
+
+		it('can retrieve blocks with transactions at heights', () => {
```
Jaguar0625, line 447: consider grouping these in own describe

rest/src/db/CatapultDb.js
Jaguar0625: don't you only need timestamp?

rest/test/db/CatapultDb_spec.js
Jaguar0625: consider swapping order of tests so they go from simpler to more complex

rest/src/db/CatapultDb.js
fboucquez: > why this projection?

To keep it consistent with the singular `blockAtHeight` 
https://github.com/symbol/catapult-rest/blob/e730a5ea4d8c25fc4aedf534962de4040eef3c83/rest/src/db/CatapultDb.js#L220. 

> don't you only need timestamp?

In this case yes, for #637 fix I would need the feeMultipler. 

Having said that, I'll add the custom projection.

rest/test/plugins/receipts/ReceiptsDb_spec.js
fboucquez: I was testing the unique ids, it's tested elsewhere so rolling this back

catapult-sdk/test/plugins/receipts_spec.js
fboucquez: Ok, I see, thanks. Would we want to add timestamp to `addressResolutionStaments` and `mosaicResolutionStatements`? 

To these endpoints?

http://ngl-dual-101.testnet.symboldev.network:3000/statements/resolutions/address
http://ngl-dual-101.testnet.symboldev.network:3000/statements/resolutions/mosaic

Or separate transactionStements from the other 2 is fine?

fboucquez:
@Jaguar0625 , do you like the location of the timestamp next to id? I'm not sure where is the best place, top-level timestamp, inside statement, in a meta object/field, or in a (partial) block object/field.

Jaguar0625:
> @Jaguar0625 , do you like the location of the timestamp next to id? I'm not sure where is the best place, top-level timestamp, inside statement, in a meta object/field, or in a (partial) block object/field.

i think in meta would probably be preferred. @gimre-xymcity do you have any opinion?

rest/test/db/CatapultDb_spec.js
fboucquez: Added describe, a new projection unit test and resorted tests.

rest/src/db/CatapultDb.js
fboucquez: Added custom projection + unit tests

fboucquez:
Remaining questions:
1 - Where to put the timestamp in the response?
2 - Should we add the timestamp to mosaic and address statements for consistency?

gimre-xymcity:
> > @Jaguar0625 , do you like the location of the timestamp next to id? I'm not sure where is the best place, top-level timestamp, inside statement, in a meta object/field, or in a (partial) block object/field.
> 
> i think in meta would probably be preferred. @gimre-xymcity do you have any opinion?

meta is probably most appropriate

fboucquez:
@Jaguar0625 , I've moved the timestamp to the meta field and I have added the meta/timestamp to address and mosaic statements. 

image

catapult-sdk/test/plugins/receipts_spec.js
Jaguar0625: either is fine

rest/src/plugins/receipts/ReceiptsDb.js
```diff
@@ -109,7 +111,41 @@ class ReceiptsDb {
...
+		);
+		const blocks = await this.catapultDb.blocksAtHeights(blockHeights,
+			{ 'block.timestamp': 1, 'block.height': 1 });
+		const data = page.data.map(pageItem => {
+			const statementBlock = blocks.find(
+				blockInfo => blockInfo.block.height.equals(
+					pageItem.statement.height
+				)
+			);
+			if (!statementBlock) {
```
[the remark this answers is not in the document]
fboucquez: Thanks for the explanation @Jaguar0625 . 

I considered the index map but my thinking was, the max page size is 100 (default is 20), in general, the worst case would be 100 individual blocks. The height field is a Long object, I cannot index by height as-is. I would need to create a Map<string, Block> where the index/key is a string representation of the Long (probably decimal of hex representation). The cost of converting the heights to string (to create the map and when looking by height index) could be worse than just the O(nlog(n)) of 100 elements, plus the memory cost of the temp map.

Do you have a better way of doing the index map? I can change it if you like.

rest/test/db/CatapultDb_spec.js
```diff
@@ -445,6 +443,72 @@ describe('catapult db', () => {
 		});
 	});
 
+	describe('', () => {
```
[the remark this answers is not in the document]
fboucquez: Uff, missed that. I'll add

rest/src/plugins/receipts/ReceiptsDb.js
```diff
@@ -109,7 +111,41 @@ class ReceiptsDb {
...
+				throw new Error(
+					`Cannot find block with height ${pageItem.statement.height.toString()}`
+				);
+			}
+			if (!statementBlock.block.timestamp) {
+				throw new Error(
+					`Cannot find timestamp in block with height ${pageItem.statement.height.toString()}`
+				);
+			}
+			return { meta: { timestamp: statementBlock.block.timestamp }, ...pageItem };
```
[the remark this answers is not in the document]
fboucquez: Although it's not expected the statements to have meta right now, I've added the original meta + a comment clarifying this.

catapult-sdk/test/plugins/receipts_spec.js
fboucquez: https://github.com/symbol/symbol-openapi/issues/279 for the open api change

rest/src/plugins/receipts/ReceiptsDb.js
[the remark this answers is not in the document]
Jaguar0625: up to @gimre-xymcity

Jaguar0625 (approved):

rest/src/plugins/receipts/ReceiptsDb.js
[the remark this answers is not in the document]
fboucquez: @gimre-xymcity , are happy about the answer? Do you want the list to be indexed?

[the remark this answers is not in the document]
gimre-xymcity: don't mind

PR #656, opened 2021-08-30
fix: rolledback zeromq 5.2.8 > 5.2.0

fboucquez:
The last `yarn upgrade` upgraded zeromq from  5.2.0 to 5.2.8. This broke the alpine image. Rest was crashing with segmentation fault errors (without any log took me a while to figure the error out).

This PR rolls back the zeromq locking up the old version to fix the alpha images.

Moving forward, we can either find the missing dockerfile's apk depdency or just change the image to ubuntu that works.

An ubuntu image is bigger than alpine but we are already using ubuntu for other images so if the same tag is used, we may be saving space.

Zeromq releases
https://github.com/zeromq/zeromq.js/releases


yilmazbahadir (dismissed):
Liaised with @fboucquez on the issue today, agreed on the version locking to `5.2.0` for the time being. Created the issue https://github.com/symbol/catapult-rest/issues/657 for us to place a fix and unlock the version.

Jaguar0625:
do you know _why_ it broke the alpine image?

yilmazbahadir:
> do you know _why_ it broke the alpine image?

https://github.com/zeromq/zeromq.js/releases/tag/v5.2.7
> Reduce the size of the package by only including prebuilds for the most used Node and Electron versions

I guessed the version quoted above is the one they introduced prebuilds but tried v5.2.6 and that one failed too, so it must be between 5.2.0 and 5.2.6. It appears they've released 8 patches in 4 days => https://github.com/zeromq/zeromq.js/releases

yilmazbahadir:
superseded by https://github.com/symbol/catapult-rest/pull/659
should we close?

Jaguar0625:
yes, you can close this one 👍

PR #658, opened 2021-09-01
feat: block's timestamp and feeMultiplier in transactions

fboucquez:
Fixes https://github.com/symbol/catapult-rest/issues/637. 

It adds blocks' timestamp and feeMultiplier to the transactions endpoints (search, multiple, and single results by id/hash). It also adds timestamp and feeMultiplier to inner transactions meta (unsure if this is ok or not).

image

 
image

### DIFF
--- a/rest/src/db/CatapultDb.js
+++ b/rest/src/db/CatapultDb.js
@@ -420,7 +420,7 @@ class CatapultDb {
 					dependentDocuments.forEach(dependentDocument => {
 						if (!document.transaction.transactions)
 							document.transaction.transactions = [];
-						if (document.meta.timestamp && document.meta.feeMultiplier) {
+						if (document.meta.timestamp !== undefined && document.meta.feeMultiplier !== undefined) {
 							dependentDocument.meta.timestamp = document.meta.timestamp;
 							dependentDocument.meta.feeMultiplier = document.meta.feeMultiplier;
 						}
@@ -432,25 +432,39 @@ class CatapultDb {
 			})));
 	}
 
+	addBlockMetaToTransactionList(list) {
+		return this.addBlockMetaToEntityList(list, ['timestamp', 'feeMultiplier'],
+			item => item.meta.height);
+	}
+
 	/**
-	 * It retrieves and adds the blocks information to the transactions' meta.
+	 * It retrieves and adds the blocks information to the entities' meta.
 	 *
 	 * The block information includes its timestamp and feeMultiplier
 	 *
-	 * @param {object[]} list the list without the added block information.
+	 * @param {object[]} list the entity list without the added block information.
+	 * @param {string[]} fields the list of fields to be be copied from the block's to the entity's meta.
+	 * @param {Function} getHeight a function that returns the block's height of a given entity.
 	 * @returns {Promise<object[]>} this list with the added block information.
 	 */
-	async addBlockMetaToTransactionList(list) {
-		const isValidHeight = height => height && '0' !== height.toString();
+	async addBlockMetaToEntityList(list, fields, getHeight) {
+		const isValidHeight = height => height && 0 !== height.toInt();
 
 		const blockHeights = uniqueLongList(
-			list.map(item => item.meta.height).filter(isValidHeight)
+			list.map(item => getHeight(item)).filter(isValidHeight)
 		);
-		const blocks = await this.blocksAtHeights(blockHeights,
-			{ 'block.timestamp': 1, 'block.height': 1, 'block.feeMultiplier': 1 });
+
+		const projection = {
+			'block.height': 1,
+			...fields.reduce((acc, field) => {
+				acc[`block.${field}`] = 1;
+				return acc;
+			}, {})
+		};
+		const blocks = await this.blocksAtHeights(blockHeights, projection);
 
 		return list.map(item => {
-			const { height } = item.meta;
+			const height = getHeight(item);
 			if (!isValidHeight(height))
 				return item;
 			const block = blocks.find(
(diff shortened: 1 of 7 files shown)


rest/src/db/CatapultDb.js
```diff
@@ -423,6 +432,57 @@ class CatapultDb {
...
+	/**
+	 * It retrieves and adds the blocks information to the transactions' meta.
+	 *
+	 * The block information includes its timestamp and feeMultiplier
+	 *
+	 * @param {object[]} list the list without the added block information.
+	 * @returns {Promise<object[]>} this list with the added block information.
+	 */
+	async addBlockMetaToTransactionList(list) {
+		const isValidHeight = height => height && '0' !== height.toString();
```
Jaguar0625, line 444: 1. why do you need this check?
2. why are you comparing strings?

rest/src/db/CatapultDb.js
```diff
@@ -423,6 +432,57 @@ class CatapultDb {
...
 
+	/**
+	 * It retrieves and adds the blocks information to the transactions' meta.
+	 *
+	 * The block information includes its timestamp and feeMultiplier
+	 *
+	 * @param {object[]} list the list without the added block information.
+	 * @returns {Promise<object[]>} this list with the added block information.
+	 */
+	async addBlockMetaToTransactionList(list) {
```
Jaguar0625, line 443: i think in previous review (for receipts) you used variable name page. is there a reason you are using list here instead?

rest/src/db/CatapultDb.js
```diff
@@ -423,6 +432,57 @@ class CatapultDb {
 			})));
 	}
 
+	/**
```
Jaguar0625, line 435: this is quite similar to addBlockMeta (except that it is also adding feeMultiplier). consider refactoring the two functions, if possible.

rest/src/db/CatapultDb.js
```diff
@@ -401,20 +401,29 @@ class CatapultDb {
...
 
 				return this.queryDependentDocuments(collectionName, [document.id]).then(dependentDocuments => {
 					dependentDocuments.forEach(dependentDocument => {
 						if (!document.transaction.transactions)
 							document.transaction.transactions = [];
-
+						if (document.meta.timestamp && document.meta.feeMultiplier) {
+							dependentDocument.meta.timestamp = document.meta.timestamp;
+							dependentDocument.meta.feeMultiplier = document.meta.feeMultiplier;
+						}
```
Jaguar0625, line 426: this does not look right since 0 is valid value for both?
also, is there any way only one of these could be set?

rest/test/db/CatapultDb_spec.js
```diff
@@ -695,23 +696,63 @@ describe('catapult db', () => {
...
 
 		return transactions;
 	};
 
+	/**
+	 * It creates the blocks for the given transactions.
+	 *
+	 * @param {object[]} transactions the transactions to know the required blocks
+	 * @returns {object[]} the block of the transactions to be stored
+	 */
```
Jaguar0625, line 726: generally, we don't need to fully document private helper functions

rest/src/db/CatapultDb.js
fboucquez: 1. For partial and unconfirmed transactions where the stored block is Long(0)
2. The stored height is actually Mongo's Long value, the unit tests stored just regular numbers. I have updated the unit tests so I can swap toString() to toInt()

rest/src/db/CatapultDb.js
fboucquez: Receipt only has paginated searches, transaction has paginated searches, and by ids/hashes lists. I can swap Recipient to use lists and (re)construct the page in the main method.

rest/src/db/CatapultDb.js
fboucquez: Good catch, the timestamp would pass Long(0) but I think feeMultipler is a regular number. I'll update the check for undefined checking

rest/src/db/CatapultDb.js
fboucquez: So I had a second look and it wasn't that bad. The new commit shares the `addBlockMetaToEntityList` method.

rest/test/db/CatapultDb_spec.js
fboucquez: Cool, public `addBlockMetaToEntityList` has the doc as it's used externally too

rest/src/db/CatapultDb.js
Jaguar0625: nice job on the refactoring 👍

rest/src/db/CatapultDb.js
```diff
@@ -423,6 +432,65 @@ class CatapultDb {
 			})));
 	}
 
+	addBlockMetaToTransactionList(list) {
+		return this.addBlockMetaToEntityList(list, ['timestamp', 'feeMultiplier'],
+			item => item.meta.height);
```
Jaguar0625, line 437: if this doesn't fit on one line, we usually drop all params on their own lines

rest/src/db/CatapultDb.js
```diff
@@ -423,6 +432,65 @@ class CatapultDb {
 			})));
 	}
 
+	addBlockMetaToTransactionList(list) {
```
Jaguar0625, line 435: you might want to document this?

Jaguar0625 (approved):

rest/src/db/CatapultDb.js
Jaguar0625: that answer makes sense, so can leave as is

rest/src/db/CatapultDb.js
fboucquez: one liner now

rest/src/db/CatapultDb.js
fboucquez: documented

Jaguar0625 (approved):

PR #659, opened 2021-09-09
fix: docker image base migrated to ubuntu v20.04 and zeromq version upgraded to v5.2.8

### LINKED ISSUE #657: Unlock zeromq version and fix the missing dependency issue in the node alpine image
Currently, `zeromq`  dependency is locked to version `5.2.0` as the latest version `5.2.8` introduces some dependencies missing in the `node:12.18.1-alpine` image thus causing `rest` container to crash (err: segmentation fault).

We should be adding the missing dependencies to the dockerfile (using node-alpine image) and unlocking the `zeromq` version.

p.s. using node-ubuntu version could bring a memory overhead thus ideally should be avoided.

yilmazbahadir:
fixes #657 

* in order to use the same base image(_ubuntu:20.04_) across all images(to save disk space), had to write a custom docker file since [the official docker node releases](https://hub.docker.com/_/node) are depending on the debian instead of ubuntu(_which is also based on debian_ but the docker images are unrelated thus no reuse).
    * https://github.com/nodejs/docker-node/tree/dc340d0bf2119dee534106ef012e85861cda8b84#image-variants
    * https://github.com/tianon/docker-brew-ubuntu-core/blob/49f002ba206e2cea2024aaa9f6f4ee4e9fb5c084/focal/Dockerfile
* kept the same node version in the previous docker file(node v12.18.1)
* tested locally ✓ 
     * run a local bootstrap(w/ testnet preset and local rest image), connected local wallet to the local node(screencast 👇 )
     https://www.loom.com/share/a69ee04c31b849d8bf922c96e775afdf
     * run a script using SDK block listener, result 👇 
     
image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,23 @@
-FROM node:12.18.1-alpine
-RUN apk add --update\
- python \
- python3 \
- build-base \
- zeromq-dev \
- && rm -rf /var/cache/apk/*
+FROM ubuntu:20.04
+
+ENV NODE_VERSION=12.18.1
+ENV NVM_DIR=/root/.nvm
+
+RUN mkdir -p ${NVM_DIR}
+RUN apt-get update && apt-get install -y curl
+
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh | bash
+RUN . "$NVM_DIR/nvm.sh" && nvm install ${NODE_VERSION}
+RUN . "$NVM_DIR/nvm.sh" && nvm use v${NODE_VERSION}
+RUN . "$NVM_DIR/nvm.sh" && nvm alias default v${NODE_VERSION}
+ENV PATH="${NVM_DIR}/versions/node/v${NODE_VERSION}/bin/:${PATH}"
+RUN node --version
+RUN npm --version
+
+RUN npm install -g yarn
+
 WORKDIR /app
 COPY . /app/catapult-rest
 RUN cd catapult-rest \
- && ./yarn_setup.sh
-RUN cd catapult-rest/rest
+    && ./yarn_setup.sh
 WORKDIR /app/catapult-rest/rest
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,5 @@ RUN npm install -g yarn
 
 WORKDIR /app
 COPY . /app/catapult-rest
-RUN cd catapult-rest \
-    && ./yarn_setup.sh
+RUN cd catapult-rest && ./yarn_setup.sh
 WORKDIR /app/catapult-rest/rest
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM ubuntu:20.04
 
 ENV NODE_VERSION=12.18.1
-ENV NVM_DIR=/root/.nvm
+ENV NVM_DIR=/usr/local/.nvm
 
 RUN mkdir -p ${NVM_DIR}
 RUN apt-get update && apt-get install -y curl
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,11 @@
 FROM ubuntu:20.04
 
-ENV NODE_VERSION=12.18.1
-ENV NVM_DIR=/usr/local/.nvm
-
-RUN mkdir -p ${NVM_DIR}
-RUN apt-get update && apt-get install -y curl
-
-RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh | bash
-RUN . "$NVM_DIR/nvm.sh" && nvm install ${NODE_VERSION}
-RUN . "$NVM_DIR/nvm.sh" && nvm use v${NODE_VERSION}
-RUN . "$NVM_DIR/nvm.sh" && nvm alias default v${NODE_VERSION}
-ENV PATH="${NVM_DIR}/versions/node/v${NODE_VERSION}/bin/:${PATH}"
-RUN node --version
-RUN npm --version
-
-RUN npm install -g yarn
+RUN apt-get update && apt-get install -y curl \
+    && curl -fsSL https://deb.nodesource.com/setup_12.x | bash - \
+    && apt-get install -y nodejs \
+    && node --version \
+    && npm --version \
+    && npm install -g yarn
 
 WORKDIR /app
 COPY . /app/catapult-rest
--- a/rest/package.json
+++ b/rest/package.json
@@ -15,7 +15,7 @@
     "lint:fix": "eslint src test --fix",
     "lint:jenkins": "eslint -o tests.catapult.lint.xml -f junit src test || exit 0",
     "bootstrap-clean": "symbol-bootstrap clean",
-    "bootstrap-start": "symbol-bootstrap start -a light -c bootstrap-preset.yml --healthCheck --noPassword",
+    "bootstrap-start": "symbol-bootstrap start -a light -c bootstrap-preset-local.yml --healthCheck --noPassword",
     "bootstrap-start-testnet": "symbol-bootstrap start -p testnet -a dual -c bootstrap-preset-testnet.yml --healthCheck --noPassword",
     "bootstrap-start-mainnet": "symbol-bootstrap start -p testnet -a dual -c bootstrap-preset-mainnet.yml --healthCheck --noPassword",
     "bootstrap-start-detached": "symbol-bootstrap start -a light -c bootstrap-preset.yml --detached --healthCheck --noPassword",
@@ -57,6 +57,6 @@
     "sshpk": "1.16.1",
     "winston": "^3.2.1",
     "ws": "^7.1.0",
-    "zeromq": "^5.1.0"
+    "zeromq": "^5.2.8"
   }
 }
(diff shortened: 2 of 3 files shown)


fboucquez (dismissed):

Dockerfile
```diff
@@ -1,13 +1,23 @@
...
+RUN npm install -g yarn
+
 WORKDIR /app
 COPY . /app/catapult-rest
 RUN cd catapult-rest \
- && ./yarn_setup.sh
-RUN cd catapult-rest/rest
-WORKDIR /app/catapult-rest/rest
+    && ./yarn_setup.sh
+WORKDIR /app/catapult-rest/rest
```
Jaguar0625, line 23: fix line ending

Dockerfile
```diff
@@ -1,13 +1,23 @@
...
-RUN apk add --update\
- python \
- python3 \
- build-base \
- zeromq-dev \
- && rm -rf /var/cache/apk/*
+FROM ubuntu:20.04
+
+ENV NODE_VERSION=12.18.1
+ENV NVM_DIR=/usr/local/.nvm
```
Jaguar0625, line 4: i don't _mind_ NVM (and it's great for development), but not sure if you need it since this will only ever need exactly one node version

Dockerfile
```diff
@@ -1,13 +1,23 @@
...
+
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh | bash
+RUN . "$NVM_DIR/nvm.sh" && nvm install ${NODE_VERSION}
+RUN . "$NVM_DIR/nvm.sh" && nvm use v${NODE_VERSION}
+RUN . "$NVM_DIR/nvm.sh" && nvm alias default v${NODE_VERSION}
+ENV PATH="${NVM_DIR}/versions/node/v${NODE_VERSION}/bin/:${PATH}"
+RUN node --version
+RUN npm --version
+
+RUN npm install -g yarn
```
Jaguar0625, line 17: each of these RUNs creates a separate docker layer, which will lead to _some_ bloat. nothing is particularly wrong with it, but it's probably not what you want.

@gimre-xymcity or @Wayonb can probably give more details about this.

Jaguar0625 (changes requested):

Dockerfile
fboucquez: Good catch about nvm.

We can just install the node distro without nvm, that's useful for dev but not production.

https://github.com/nodesource/distributions/blob/master/README.md

yilmazbahadir: The code you shared would download the latest `12.x` version not `12.18.1` so the `nvm` choice was intentional, to be able to specify the exact version `12.18.1` (since didn't feel safe to change the `node` version at this stage). I also didn't like the idea of having `nvm` in production, but during my research saw that it was the common way to handle it smoothly since it abstracts the installing process. If we want to have the latest `12.x` version, I'll be happy to change it.

Dockerfile
yilmazbahadir: Please see the comment below, it was intentional to use a particular(12.18.1) version.

Dockerfile
yilmazbahadir: fixed.

Dockerfile
fboucquez: Ofc, fix version means explicit version upgrades. A bit more overhead to maintain but safer.

fboucquez: Yet another comment, travis uses the latest 12 every build. 

https://github.com/symbol/catapult-rest/blob/dev/.travis.yml#L4

So we are unit testing ~100% coverage with the latest node 12 version.

yilmazbahadir: I'm keen to upgrade it to the latest 12.x tbh, just checked the [nodejs 12.x change log](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md) between 12.18.1 and 12.2.6(latest) seen a lot of security fixes which is good. I assume it's safe to upgrade but we'll need proper testing and some soak time in testnet before mainnet release. 
What do you think @Jaguar0625 and @Wayonb ?
1. Upgrade it to the nodejs latest version(12.x) in the docker file (which will have the latest 12.x in each build)
2. keep nodejs version fixed at 12.18.1 (as it is)

Dockerfile
```diff
@@ -1,13 +1,22 @@
...
+RUN mkdir -p ${NVM_DIR}
+RUN apt-get update && apt-get install -y curl
+
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh | bash
+RUN . "$NVM_DIR/nvm.sh" && nvm install ${NODE_VERSION}
+RUN . "$NVM_DIR/nvm.sh" && nvm use v${NODE_VERSION}
+RUN . "$NVM_DIR/nvm.sh" && nvm alias default v${NODE_VERSION}
+ENV PATH="${NVM_DIR}/versions/node/v${NODE_VERSION}/bin/:${PATH}"
+RUN node --version
+RUN npm --version
```
gimre-xymcity, line 15: not sure if you want it here at all, and if you do, can you move it into single `RUN` with npm?
```
RUN node --version
    && npm --version
    && npm install -g yarn
```

yilmazbahadir: Thanks @gimre-xymcity, already addressed that concern, it'll look like the following if we choose to upgrade it to `v12.x (latest)`. Any ideas on this => https://github.com/symbol/catapult-rest/pull/659#discussion_r707290619 ?
```
FROM ubuntu:20.04

RUN apt-get update && apt-get install -y curl \
    && curl -fsSL https://deb.nodesource.com/setup_12.x | bash - \
    && apt-get install -y nodejs \
    && node --version \
    && npm --version \
    && npm install -g yarn

WORKDIR /app
COPY . /app/catapult-rest
RUN cd catapult-rest && ./yarn_setup.sh
WORKDIR /app/catapult-rest/rest

```

gimre-xymcity: #perfect 👍

Dockerfile
gimre-xymcity: Can't really help with that, but would _probably_ use `fixed` version

Dockerfile
Jaguar0625: yes, what you pasted in the comment looks good 👍

yilmazbahadir: just to make sure @Jaguar0625, the code pasted above will install the `nodejs v12.x (latest)` in each build(it won't be a fixed version). Version upgrades will be mostly security patches and npm upgrades but there is a risk of course.  I'm going to push the changes if we are all okay with using the latest version, 12.x.

yilmazbahadir: Okay, found a workaround to pin the version, please see the dockerfile content below.
```
FROM ubuntu:20.04

RUN apt-get update && apt-get install -y curl \
    && curl -o nodejs.deb https://deb.nodesource.com/node_12.x/pool/main/n/nodejs/nodejs_12.18.1-deb-1nodesource1_amd64.deb \
    && apt-get install -y ./nodejs.deb \
    && rm nodejs.deb \
    && rm -rf /var/lib/apt/lists/* \
    && node --version \
    && npm --version \
    && npm install -g yarn

WORKDIR /app
COPY . /app/catapult-rest
RUN cd catapult-rest && ./yarn_setup.sh
WORKDIR /app/catapult-rest/rest
```
Local tests worked fine. If everyone aggrees I'll push the changes using the fixed version?

fboucquez: Nice finding. Fix version without nvm.

I'm ok using the old version although it's a good opportunity to upgrade the latest fix version 12.2.6

Jaguar0625: i don't really see a compelling reason to use an older version? @0x6861746366574?

yilmazbahadir: Hey @0x6861746366574 , we need your input on this one, could you please share your view on this?
Current nodejs (in catapult-rest on the symbol network) version is `12.18.1`. 
We have 3 options:
1. Keep the version as it is as `12.18.1`. https://github.com/symbol/catapult-rest/pull/659#discussion_r710436996
2. Upgrade the version to `12.2.6` and pin it.
3. Use the latest `12.x` version on each build (currently the latest version is `12.22.6` but will probably change in the future, security patches, npm upgrades, etc.) https://github.com/symbol/catapult-rest/pull/659#discussion_r709508208

My vote would be on option 2 as well (with enough soak time in testnet ofc)

0x6861746366574: **Option 3.** 

One of the biggest changes in cryptocurrency compared to traditional software development is that you want to make sure your dependencies are always operating on the latest version. The attack vectors for chains are very different compared to traditional financial systems - and the prize is far higher, too. 

There is additional burden on us and operators with frequent upgrades, yes, but, that's the entire point of network fees: ongoing reward for ongoing upkeep.

fboucquez (approved):

Dockerfile
yilmazbahadir: Updated the dockerfile with the latest nodejs version as @0x6861746366574  suggested. 

Unfortunately, Travis is failing due to coveralls outage for some time now. We can wait until the service recovers or continue to merge.

> /home/travis/build/symbol/catapult-rest/catapult-sdk/node_modules/coveralls/bin/coveralls.js:19
      throw err;
      ^
Bad response: 503 <h2>This website is under heavy load (queue full)</h2><p>We're sorry, too many people are accessing this website at the same time. We're working on this problem. Please try again later.</p>

image

Jaguar0625 (approved):

PR #683, opened 2022-01-06
fix: removed decoded address

AnthonyLaw:
### Current behavior 
- All the addresses in the REST response are in decoded format. 
- Example `98E0D138EAF2AC342C015FF0B631EC3622E8AFFA04BFCC56`

### What was the issue?
- `Decoded` address format is not user-friendly. We have feedback from the community and discussion, change the address to encoded format.
- Example `TDQNCOHK6KWDILABL7YLMMPMGYRORL72AS74YVQ`

### What's the fix?
- Convert addresses from decoded to encoded (base32) format.
- updated unit test

### DIFF
--- a/rest/src/db/dbUtils.js
+++ b/rest/src/db/dbUtils.js
@@ -82,12 +82,11 @@ const dbUtils = {
 	},
 
 	/**
-     * Formats binary to a base32 address or hex address
+     * Formats binary to a base32 address
      * @param {MongoDb.Binary} binary Address|NamespaceId from MongoDb.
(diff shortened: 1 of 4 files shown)


0x6861746366574 (approved):

0x6861746366574:
@AnthonyLaw Before merging, could you adjust the issue to follow this format [seen here](https://github.com/symbol/statistics-service/pull/121)? 

As a rule of thumb, you want to describe these three things:

* **What** the current behaviour is.
* **Why** this is an issue.
* **How** you changed this behaviour. 

I know it's a bit overkill for this change, but we want to get in a habit of doing this for each PR/issue.

